PR #7, opened 2016-12-06
Use npm-run-all to execute scripts

hetec:
Just a small addition to make the build still a bit simpler


flashback2k14:
you have added the node_modules folder!

hetec:
ahhh sorry

PR #8, opened 2016-12-06
#4 - use npm-run-all to execute multiple scripts

hetec:


hetec:
So diesmal ohne node-modules ;)

PR #9, opened 2016-12-08
#6 - bootstrap database

flashback2k14:
see the readme :-)


server/src/routes/api/v1/users/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  constructor (userService: UserService, router: Router) {
+    this._userService = userService;
+    this._router = router;
+  }
+
+  createRoutes (): Router {
+    // get all users
+    this._router.get("/", (req: Request, res: Response, next: NextFunction) => {
+      this._userService.getAllUsers()
+        .then((result: SocoboUser[]) => res.status(200).json(result))
```
hetec: I think we should create a common object to provide the status codes 200, 400 ... instead of using these magic numbers

server/src/models/socobouser.ts
```diff
@@ -0,0 +1,5 @@
+export class SocoboUser {
```
hetec: I would suggest to not use the Socobo-Prefix but I think thats the fault of postgres, since User is a protected key word or so!?

server/src/logic/services/user.service.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export class UserService {
+  private _db: IDatabase<any>;
+
+  constructor (db: IDatabase<any>) {
+    this._db = db;
+  }
+
+  getAllUsers (): Promise<SocoboUser[]> {
+    return new Promise((resolve, reject) => {
```
hetec: I think you don't need to create a new Promise because the many function already returns a promise

server/test/user.route.v1.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  it(`route /api/v1/users/1 should return one user with 
+      id, username, password property`, () => {
+        chai.request(Server).get("/api/v1/users/1")
+          .then(res => {
+            expect(res.body).to.haveOwnProperty("id");
+            expect(res.body).to.haveOwnProperty("username");
+            expect(res.body).to.haveOwnProperty("password");
+          }); 
+  });
+});
```
hetec: The tests seem to pass correctly but I always get this error:

/Users/patrick/dev/socobo/SocoboPlusPlus/server/node_modules/ts-node/dist/index.js:160
throw new TSError(formatDiagnostics(diagnosticList, cwd, ts, lineOffset));
^
TSError: ⨯ Unable to compile TypeScript
src/logic/services/user.service.ts (2,10): Module '"/Users/patrick/dev/socobo/SocoboPlusPlus/server/src/models/socobouser"' has no expor
ted member 'SocoboUser'. (2305)

server/src/routes/api/v1/users/index.ts
```diff
@@ -0,0 +1,32 @@
+import { Router, Request, Response, NextFunction } from "express";
+import { UserService } from "./../../../../logic/services/user.service";
+import { SocoboUser } from "./../../../../models/socobouser";
+
+export class UsersRouteV1 {
+  private _userService: UserService;
+  private _router: Router; 
+
+  constructor (userService: UserService, router: Router) {
```
hetec: You could use the short cut version of the constructor

server/src/models/socobouser.ts
flashback2k14: Yes. User is a keyword in PostgreSQL.

server/test/user.route.v1.test.ts
flashback2k14: Hmm. I haven't get this!?

server/src/routes/api/v1/users/index.ts
hetec: I Just called them magic Numbers because they are hard coded. Ok maybe this is not completly true because status codes have actually a meaning. But I think named variables would be slightly more readable. Because I not always know what the status code means if i only see the number

server/test/user.route.v1.test.ts
hetec: Ok i try this

server/src/routes/api/v1/users/index.ts
flashback2k14: Okay. Add this class to this PR or the Refactoring PR?

flashback2k14: it's merged!

server/test/user.route.v1.test.ts
hetec: fyi the error occurs if some other process is already listening to the defined port

flashback2k14: Ahh okay. Watch test changes also not working. I haven't figured how to close the server after Tests. Maybe we need to provide a function to close the server.

server/src/routes/api/v1/users/index.ts
flashback2k14: I have added a object for error messages.

server/src/models/backenderror.ts
```diff
@@ -0,0 +1,9 @@
+export class BackendError {
```
hetec: Nice idea! But I would not propagate the whole stacktrace to the client. Maybe we should log the errors to a errorlog and only provide a clear message to the client. Additionally we should collect the messages in a central location (object, file etc.)

I would open a separate issue for that

PR #15, opened 2016-12-11
#1 - bootstrap frontend application

flashback2k14:
angular-cli must be global

```bash
npm install -g angular-cli
```


flashback2k14:
@lyio @hetec a Short reminder! 😀

hetec:
I will look at this pr tomorrow

hetec:
Looks good at the first glance and runs fine! On thing I don't really understand is why we don't need to use system js as module loader? To my understanding it's because we focus on ES6. Is this correct? If this is the point, we should maybe introduce system js to the project at some point to support older browsers.

flashback2k14:
I think angular-cli and the crazy underlying webpack stuff makes the transpiling to ES5. SystemJS was needed with angular2 Beta releases.

PR #17, opened 2016-12-20
Add information directly to the error

hetec:
Add the error object with the contained information to the in memory log and to the log file. So the logging service is independent from the error format and can handle different error types. 

We would only need to create the Error types we need (like the dbError) with additional properties (like the query)

What do you think, makes this sense?


hetec:
The tests are a bit dangerous since there is no way to replace the log cache with one only for the tests. That means the test operate on the real log cache. But since everything is static an shared state there is no way to fix this. On the other hand we would need to inject the LogService in each object which uses it if we would remove the statics from the class. This would be very inconvenient. So I would think we should keep this implementation

flashback2k14:
Hmm, what if we add a clear function? Or add a namespace or something for dev, prod and test?

flashback2k14:
I don't know if I get your point!?

hetec:
The Problem is that the clear function would also delete existing log messages which are not produced by the test. But maybe that is not so bad since it just a cache.

If you have some siggestions to make the Tests mit independent,  your changes are welcome.

flashback2k14:
You know I'm not the testing pro, but I look into it. Maybe I found same ways to split dev/prod and test.

flashback2k14:
i think i have solved the problem. i replace the winston logger with a spy logger. Only the server test write into a log file, but this is not the prod / dev log file. Please look at my commits. Sorry for the many changes. :-)

Actually we have no problem with log messages that are stored inside the Array from prod or dev inside the tests, because they live only in the memory and the test suite restarts the server.

PR #24, opened 2016-12-24
#19 - User Login

flashback2k14:
now it's possible to login with username or email, create login function for user route tests

### DIFF
--- a/server/src/logic/services/auth.service.ts
+++ b/server/src/logic/services/auth.service.ts
@@ -3,16 +3,26 @@ import { UserService } from "./user.service";
 import { CryptoUtils } from "./../utils/cryptoUtils";
 import { ErrorUtils } from "./../utils/errorUtils";
 import { Config } from "./../../config";
-import { ApiError, DbError, SocoboUser } from "./../../models/index";
+import { 
+  ApiError, DbError, SocoboUser, LoginResult 
+} from "./../../models/index";
 
 
 export class AuthService {
   constructor (private _userService: UserService) {}
 
-  login (email: string, password: string): Promise<Object> {
+  login (isEmailLogin: boolean, usernameOrEmail: string, password: string): Promise<LoginResult> {
     return new Promise((resolve, reject) => {
-      // search for the user by provided email
-      this._userService.getUserByEmail(email)
+      // create promise holder
+      let loginPromise: Promise<SocoboUser>;
+      // search for the user by provided email or username
+      if (isEmailLogin) {
+        loginPromise = this._userService.getUserByEmail(usernameOrEmail);
+      } else {
+        loginPromise = this._userService.getUserByUsername(usernameOrEmail);
+      }
+      // login the user or reject
+      loginPromise
         .then((user: SocoboUser) => {
           // check if the user was found
           if (!user) {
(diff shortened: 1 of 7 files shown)


server/src/logic/services/auth.service.ts
```diff
@@ -3,16 +3,26 @@ import { UserService } from "./user.service";
...
+import { 
+  ApiError, DbError, SocoboUser, LoginResult 
+} from "./../../models/index";
 
 
 export class AuthService {
   constructor (private _userService: UserService) {}
 
-  login (email: string, password: string): Promise<Object> {
+  login (isEmailLogin: boolean, usernameOrEmail: string, password: string): Promise<LoginResult> {
```
hetec: I would use two methods (loginInWithUsername and logInWithEmail) instead of the flag for the sake of readability but this would maybe blow the code up.

Or you could maybe use two parameters (username and email) and check each only if present.

server/src/routes/api/v1/auth/index.ts
```diff
@@ -1,17 +1,24 @@
...
   constructor (private _authService: AuthService, private _router: Router) {}
 
   createRoutes (): Router {
     // login the user
-    this._router.post("/login", (req: Request, res: Response, next: NextFunction) => {
-      this._authService.login(req.body.email, req.body.password)
-        .then((result: SocoboUser) => res.status(200).json(result))
+    this._router.post("/login", this.checkLoginRequest, (req: Request, res: Response, next: NextFunction) => {
+      // get data from request
+      let isEmailLogin: boolean = req.body.isEmailLogin;
```
hetec: Maybe we could do the isEmailOrUsername check in the backend. The frontend could just send one string in the body (email or username) and the backend would then first assume that the string is a email (check against email) if this fails the backend checks the provided string against the username and if this fails the login was invalid.

So we could avoid these logic in the frontend. Just an idea.

server/src/routes/api/v1/auth/index.ts
```diff
@@ -38,4 +45,24 @@ export class AuthRouteV1 {
 
     return this._router;
   }
+
+  private checkLoginRequest (req: Request, res: Response, next: NextFunction): any {
```
hetec: I think this should be done via input validation in any way. Maybe you need a model for the login

server/src/routes/api/v1/auth/index.ts
flashback2k14: Hmm. At the first this is Backend code. :-)

In the middleware I check the request body and set the isEmailLogin flag. I don't like the fail check. I think this takes too much time.

server/src/logic/services/auth.service.ts
flashback2k14: I check this in the middleware.

server/src/routes/api/v1/auth/index.ts
flashback2k14: Hmm, I like my middleware. 😛

server/src/routes/api/v1/auth/index.ts
hetec: Ok

server/src/routes/api/v1/auth/index.ts
hetec: OK, was just an idea

flashback2k14:
could we merge this? after merge i start refactoring PR #20.

PR #23, opened 2016-12-24
#22 - add database for dev and testing

flashback2k14:
Add databases for development and testing

PR #25, opened 2016-12-24
#20 - user registration

flashback2k14:
Add login and register process with email and password and protect routes.

SORRY for this big PR. :-) 

- extend SocoboUser
- extend UserService
- add AuthService
- add AuthRoute
- add CryptoUtils
- add AuthUtils
- protect API routes with AuthUtils
- integrate AuthRoute / -service into Server.ts
- restructure Server.ts
- restructure imports (first extern modules, then our modules and 2 lines between last import and export class)

Please first merge PR #23 and #24 before this should be merged!


server/src/routes/api/v1/auth/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  createRoutes (): Router {
+    // login the user
+    this._router.post("/login", (req: Request, res: Response, next: NextFunction) => {
+      this._authService.login(req.body.email, req.body.password)
+        .then((result: SocoboUser) => res.status(200).json(result))
+        .catch((error: any) => {
+          if (ErrorUtils.notFound(error)) {
+            res.status(404).json(
+                new DbError(`The user login is failed - ${error.message}!`, AuthService.name,
```
hetec: Maybe the error message should be a bit more descriptive. Does this mean that the user for the login is not registered yet?

server/src/routes/api/v1/auth/index.ts
```diff
@@ -0,0 +1,41 @@
...
+                new ApiError(`Internal Server Error: ${error.message}`, AuthService.name, 
+                              "login(email,password)", error).forResponse());
+          }
+        });
+    });
+
+    // register the user
+    this._router.post("/register", (req: Request, res: Response, next: NextFunction) => {
+      this._authService.register(req.body.email, req.body.password)
+        .then((result: SocoboUser) => res.status(200).json(result))
```
hetec: I would never return the password regardless of whether it is a hash. The client doesn't need this information. 

And I would maybe add general information about the user (no sensitive and only things like isAdmin and so) to the payload of the jwt since the jwt will be stored into a cookie, the client has always access to these information via the jwt and does not request these again and again

server/src/server.ts
```diff
@@ -71,34 +91,58 @@ class Server {
...
     let router: express.Router = express.Router();
-    // init user service
-    const userService: UserService = new UserService(this._db);
     // init and return users route
-    return new UsersRouteV1(userService, router).createRoutes();
+    return new UsersRouteV1(this._userService, router).createRoutes();
   }
 
-  private __logsRoute (): express.Router {
+  private ___logsRoute (): express.Router {
```
hetec: I really can accept the use of one _ since the tooling of js / ts is as mature as in java or so but what is the purpose of two or even three underscores??? This is really hard to read and search.

server/test/user.route.v1.test.ts
```diff
@@ -18,56 +18,107 @@ describe("User Route v1", () => {
     let user1: SocoboUser = new SocoboUser();
```
hetec: As stated above I would also prefer to have some unit tests for the single modules (auth, user service ...). These endpoint tests here are very important to test the the functionality as a whole, but they are not well suited to test each component in isolation. See also the comment above.

Since you are testing two things at a time here (the auth and the user server / user routes) you cannot easily locate the error. It could alway be one of these two components. And a change of one of these two components would alway force you to update the tests even if the other component has not changed.

.vscode/launch.json
```diff
@@ -0,0 +1,27 @@
...
+    {
+      "name": "Attach to Process",
+      "type": "node2",
+      "request": "attach",
+      "port": 5858,
+      "outFiles": [],
+      "sourceMaps": true
+    }
+  ]
+}
```
flashback2k14: Yes, but maybe you need also to debug sometimes!? But we could exclude this.

server/sql/create-tables.sql
```diff
@@ -3,5 +3,10 @@
 CREATE TABLE Socobo_User (
```
flashback2k14: Yeah, I'm not a SQL pro. Every addition is highly welcome.

server/src/logic/services/auth.service.ts
```diff
@@ -0,0 +1,96 @@
...
+import { Config } from "./../../config";
+import { ApiError, DbError, SocoboUser } from "./../../models/index";
+
+
+export class AuthService {
+  constructor (private _userService: UserService) {}
+
+  login (email: string, password: string): Promise<Object> {
+    return new Promise((resolve, reject) => {
+      // search for the user by provided email
```
flashback2k14: Yes you are right. I add comments because I hate whitespace. But these functions do too much. That's a good idea to break things down into smaller functions. I really need to read Clean code. ;-)

server/src/logic/services/auth.service.ts
```diff
@@ -0,0 +1,96 @@
...
+          // check if the provided password match the users password
+          CryptoUtils.comparePasswords(password, user.password)
+            .then((passwordMatch: boolean) => {
+              // if not match reject JWT creation
+              if (!passwordMatch) {
+                return reject(new Error("Authentication failed. Wrong password."));
+              }
+              // generate the JWT
+              jwt.sign(user, Config.TOKEN_SECRET, {
+                expiresIn: "1d"
```
flashback2k14: Good idea

server/src/logic/services/auth.service.ts
```diff
@@ -0,0 +1,96 @@
...
+        // catch some database errors
+        .catch((error: any) => {
+          // check if no data was found
+          if (ErrorUtils.notFound(error)) {
+            // hash password
+            CryptoUtils.hashPassword(password)
+              .then((hashedPassword: string) => {
+                // create new user object
+                let user: SocoboUser = new SocoboUser();
+                user.username = email.split("@")[0];
```
flashback2k14: I want provide this option on the profile page!?

.vscode/launch.json
hetec: Let us keep this, I think I will not change the editor soon

server/src/logic/utils/authUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+import { Config } from "./../../config";
+
+
+export class AuthUtils {
+
+  static checkState (req: any, res: Response, next: NextFunction) {
+    // get token from request
+    const token = req.body.token 
+                    || req.query.token 
+                    || req.headers["x-access-token"];
```
flashback2k14: Maybe add this to the config?

server/src/logic/utils/cryptoUtils.ts
```diff
@@ -0,0 +1,38 @@
+import * as bcrypt from "bcrypt";
+import { ApiError } from "./../../models/apiError";
+
+
+export class CryptoUtils {
+
+  static hashPassword (userPassword: string): Promise<string> {
```
flashback2k14: Did you mean the static thing?

server/src/logic/utils/authUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+    // check if token is available
+    if (token) {
+      // verify token
+      jwt.verify(token, Config.TOKEN_SECRET, (err: any, decoded: any)  => {
+        // error handling
+        if (err) {
```
flashback2k14: Yes, you are right.

.vscode/launch.json
flashback2k14: What editor / IDE did you use?

server/src/logic/utils/authUtils.ts
```diff
@@ -0,0 +1,42 @@
+import * as jwt from "jsonwebtoken";
+import { Router, Request, Response, NextFunction } from "express";
+import { Config } from "./../../config";
+
+
+export class AuthUtils {
+
+  static checkState (req: any, res: Response, next: NextFunction) {
```
flashback2k14: Okay. You mean your validator approach!?

server/src/logic/utils/authUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+          return res.status(400).json({success: false, message: msg});
+        } else {
+          // go to the next step
+          req.decoded = decoded;
+          next();
+        }
+      });
+    } else {
+      // return error data
+      return res.status(403).json({success: false, message: "No token provided."});
```
flashback2k14: Okay. But before I get it, I need to watch this video.

server/src/routes/api/v1/auth/index.ts
flashback2k14: Hmm, I can't remember. :-) I think yes.

server/src/routes/api/v1/auth/index.ts
flashback2k14: On PR #23 I have removed the password.

I think isAdmin is integrated into the payload with jwt.sign() the first parameter is the user object.

server/src/server.ts
flashback2k14: Sorry. :-D it shows the calling depth. :-)

server/test/user.route.v1.test.ts
flashback2k14: Okay. Could you please refactor the tests? I don't have any clue how to test this the right way. :-(

server/src/logic/utils/cryptoUtils.ts
hetec: yes

flashback2k14: Okay. I will refactor this.

.vscode/launch.json
hetec: Code

server/src/logic/services/auth.service.ts
hetec: OK thats also a good idea

.vscode/launch.json
flashback2k14: Okay. Could you please set your tabSize to 2 and if you add new dependencies to package.json please use ~ instead of ^. Then our npm installs are not to different. ~ looks for the newest patch version and ^ looks for the newest minor & patch version. Maybe its a good idea to use https://docs.npmjs.com/cli/shrinkwrap that we have the same package versions installed.

server/src/logic/utils/authUtils.ts
hetec: Yes, for instance I injected the validator into the recipe routes. But I don't want to force you to use my approach, I just think this would be good idea and it is also often suggested as good practice.

server/src/logic/utils/authUtils.ts
hetec: yes also possible

server/src/logic/utils/authUtils.ts
hetec: Okay, if you should disagree with the approach after watching the video, we can discuss this again if you want

server/src/routes/api/v1/auth/index.ts
hetec: Thats good. And yes, the whole user is included into the jwt. But we also need to remove the pw hash from the jwt since this is accessible to everyone. Maybe we should think about what goes into the jwt in general again

.vscode/launch.json
hetec: I use npm install --save so npm adds the deps. I used mostly because it allows me to use double quotes in Strings without escaping. Maybe we need some Style guide anderen linting tool

I try to consider the other things in the future

server/test/user.route.v1.test.ts
hetec: I think the Tests are OK. I would just add some unit tests. I can do this anytime later, maybe in another PR   in Order to get this one done. Do we need an issue to collect pending tests?

.vscode/launch.json
flashback2k14: you can search for npm packages inside package.json in dependencies or devDependencies. Simply start typing and strg+space and you get suggestions. Yes, i style guide is highly needed.

server/test/user.route.v1.test.ts
flashback2k14: yes, we should add the tests to an extra issue.

flashback2k14:
I think I've refactored all the discussed points. The middleware approach is really nice. i like it.
what i have done:
- rewrite login / register function to use the promise chain
- add auth middleware to AuthValidator and use Promise approach
- inject middleware and utils form server.ts into needed classes
- restructure models folder (it was really messy)
- extend Config with all JWT related stuff

server/src/routes/api/v1/auth/index.ts
```diff
@@ -0,0 +1,88 @@
...
+    
+      this._extractRequestBody(req)
+        .then((result: ExtractRequestBodyResult) => {
+          return this._authService.register(result.isEmailLogin, 
+                                            result.usernameOrEmail, 
+                                            result.password); 
+        })
+        .then((result: SocoboUser) => res.status(200).json(result))
+        .catch((error: any) => {
+          res.status(500).json(
```
hetec: Here the error problem mentioned above occurs

server/src/models/results/loginResult.ts
```diff
@@ -0,0 +1,12 @@
+import { SocoboUser } from "./../database/socoboUser";
```
hetec: Maybe I would replace Result with something like Response since these kind of object is used as response. So loginResult would become loginResponse or maybe loginResponseBody to be more concrete

server/src/routes/api/v1/auth/index.ts
```diff
@@ -0,0 +1,88 @@
+import { Router, Request, Response, NextFunction } from "express";
```
hetec: What do you think about renaming index.ts. I'm not sure why it's called index if it contains the routes. So with recipes I called this recipe.routes.ts.

server/src/models/index.ts
```diff
@@ -0,0 +1,6 @@
+export * from "./errors/apiError";
```
hetec: Is this a common pattern to reduce imports?

server/src/models/results/comparePwResult.ts
```diff
@@ -0,0 +1,6 @@
+import { SocoboUser } from "./../database/socobouser";
```
hetec: Maybe I would replace 'Result' with something like 'Response' or 'ResponseBody' since this kind of objects is mainly used as response body.

server/src/routes/api/v1/auth/index.ts
hetec: What do you think about renaming index.ts? I'm not sure why we use index if these files contain routes? With recipes I called this recipe.routes.ts

server/src/routes/api/v1/auth/index.ts
hetec: Here occurs the error problem which I have mentioned above

server/src/routes/api/v1/auth/index.ts
```diff
@@ -0,0 +1,88 @@
...
+      },
+      (req: Request, res: Response, next: NextFunction) => {
+    
+      this._extractRequestBody(req)
+        .then((result: ExtractRequestBodyResult) => {
+          return this._authService.register(result.isEmailLogin, 
+                                            result.usernameOrEmail, 
+                                            result.password); 
+        })
+        .then((result: SocoboUser) => res.status(200).json(result))
```
hetec: Maybe this should be 201 since a new account will be created?!

hetec:
After we have merged this PR I would suggest that I merge this into my recipe branch and open a PR for this as next step

server/src/logic/services/auth.service.ts
```diff
@@ -0,0 +1,122 @@
...
+
+
+export class AuthService {
+
+  constructor (
+    private _userService: UserService, 
+    private _cryptoUtils: CryptoUtils
+  ) {}
+
+  login (isEmailLogin: boolean, usernameOrEmail: string, password: string): Promise<LoginResult> {
```
flashback2k14: Yes. It's really nice now. Simpler to read and easier to understand.

server/src/logic/services/auth.service.ts
```diff
@@ -0,0 +1,122 @@
...
+            reject(error);
+          }
+        });
+    });
+  }
+
+  private _checkIfUserIsAlreadyRegistered (user: SocoboUser): Promise<any> {
+    return new Promise((resolve, reject) => {
+      if (user) {
+        return reject(new Error("Email or Username is already registered. Please use another one."));
```
flashback2k14: Yes, I'm with you. We need an error object with message and error code (BaseError) and before we return the error to the user / outside we check the error code and return the right status code and api error etc. We should add this to another issue. This is big enough.

server/src/logic/services/auth.service.ts
```diff
@@ -0,0 +1,122 @@
...
+  private _createNewUser (hashedPassword: string, usernameOrEmail: string): Promise<SocoboUser> {
+    return new Promise((resolve, reject) => {
+      if (hashedPassword.length <= 0) {
+        return reject(new Error("Hashed Password length is <= 0"));
+      }
+      let user: SocoboUser = new SocoboUser();
+      user.username = usernameOrEmail.includes("@") ? usernameOrEmail.split("@")[0] : usernameOrEmail;
+      user.email = usernameOrEmail.includes("@") ? usernameOrEmail : "";
+      user.password = hashedPassword;
+      user.image = "http://placehold.it/350x150";
```
flashback2k14: Nice catch. I will add it.

server/src/logic/utils/cryptoUtils.ts
```diff
@@ -0,0 +1,40 @@
...
+} from "./../../models/index";
+
+
+export class CryptoUtils {
+
+  hashPassword (userPassword: string): Promise<string> {
+    return new Promise((resolve, reject) => {
+      bcrypt.genSalt(10, (err, salt) => {
+        if (err) {
+          return reject(new ApiError("Error by Generating the Salt value.",
```
flashback2k14: Me too. Sounds both right for me. :-) I will change it.

server/src/models/database/socobouser.ts
```diff
@@ -0,0 +1,10 @@
+export class SocoboUser {
```
flashback2k14: Good idea. I add this in another issue. We should complete this to move forward. This PR goes into branch 12 user management and not directly into the master.

server/src/models/errors/dbError.ts
```diff
@@ -1,5 +1,6 @@
-import { LogService } from "./../logic/services/logging.service";
-import { ApiError } from "./api-error";
+import { LogService } from "./../../logic/services/logging.service";
```
flashback2k14: Yes, that's a good idea. Kebab-case for files and camel-case for classes. Add this changes to another issue.

server/src/models/index.ts
flashback2k14: Yes, I think this is a common pattern. I have seen this on many angular projects. Yes, we should add this pattern for services, Middleware, etc.

flashback2k14: I would add this to the rename issue.

server/src/routes/api/v1/auth/index.ts
flashback2k14: Yes, we could rename the routes files. This is a common practice on node. I add this to the rename issue.

flashback2k14: I add this to the rename issue.

server/src/routes/api/v1/auth/index.ts
flashback2k14: Yes, you are right. I will change this.

server/src/routes/api/v1/auth/index.ts
flashback2k14: Yes, we change this on another issue with improved error handling.

flashback2k14: Yes, we change this in another issue with improved error handling.

server/src/models/results/loginResult.ts
flashback2k14: Okay, this result goes to the outside. We could change this to your suggestions inside the rename issue.

server/src/models/results/comparePwResult.ts
flashback2k14: But this is simply a result that is only internal used.

flashback2k14:
New issues:
- improve error handling 
- renaming files and classes
- add created to SocoboUser

flashback2k14:
I think you should merge branch 12 after the renaming issue!?

flashback2k14:
could we merge this? @hetec

Issue #26, opened 2017-01-04
improve naming of files and classes


flashback2k14:
- use kebab-case for file names
- use camel-case for class names
- add middleware, service, util postfix
- add response folder to models
- rename router files

PR #30, opened 2017-01-09
#26 - improve naming of files and classes

flashback2k14:


server/src/routes/api/v1/users/users-v1.route.ts
```diff
@@ -1,13 +1,13 @@
 import { Router, Request, Response, NextFunction } from "express";
```
hetec: I wouldn't use the v1 version indicator so much because the version is already visible from the directory tree (v1 directory). I think thats redundant and makes the name harder to read

flashback2k14: Okay. Could you change this?

hetec: yes sure

flashback2k14:
LGTM

PR #32, opened 2017-01-11
27 improve error handling

hetec:
Hopefully a way to solve our error issues:

- ApiError extends Error -> We can now use ApiError or one of the classes which inherit from ApiError if we want to raise a error (e.g. in reject functions)
- ApiError takes a specific ErrorType object which describes the error, so we don't need to check the error again e.g in the routes. We can simply use the information from the ApiError (like the message or the statusCode)
- All specific ErrorType are collected in a separate file
- All error messages are also collected in a separate file, so we can change the message in one place. And if we want to support I18n we can just switch the file. The errors must not change since the message keys remain the same
- You can still use parameters in the messages (See user not found for an example)

I really hope that makes sense, if not please tell me. This error stuff is a little bit terrible and complicated.


flashback2k14:
looks really nice.

server/src/logic/services/auth.service.ts
```diff
@@ -26,7 +25,19 @@ export class AuthService {
         .then((foundUser: SocoboUser) => this._cryptoUtils.comparePasswords(password, foundUser))
         .then((compareResult: ComparePwResult) => this._validateComparePasswords(compareResult.isPasswordMatch, compareResult.user))
         .then((user: SocoboUser) => resolve(this._createLoginResult(user)))
-        .catch((error: any) => reject(error));
+        .catch((error: any) => {
+          if (ErrorUtils.notFound(error)) {
+            let e = new DbError(ERRORS.USER_NOT_FOUND);
```
flashback2k14: here is a .withArgs(usernameOrEmail) function call missing

server/src/routes/api/v1/logs/log.routes.ts
```diff
@@ -16,7 +16,7 @@ export class LogRoute {
       (req: Request, res: Response, next: NextFunction) => {
         this._authValidator.checkValidToken(req)
           .then(() => next())
-          .catch((err: ApiError) => res.status(400).json(err));
+          .catch((err: any) => res.status(err.statusCode).json(err.forReponse()));
```
flashback2k14: here is a typo in .forResponse()

server/src/error-messages.ts
```diff
@@ -0,0 +1,16 @@
+export const ERROR_MESSAGES: any = {
```
flashback2k14: i would move this into a i18n or messages folder

server/src/errors.ts
```diff
@@ -0,0 +1,80 @@
+import { ErrorType } from "./models/index"
```
flashback2k14: i would move this into the models/errors folder

server/src/errors.ts
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * First digit indicates the type:
+ * 0 => Generic errors
+ * 1 => User errors
+ * 2 => Auth errors
+ * 3 => Validation errors
+ */
+export class ERRORS {
```
flashback2k14: i would add a semicolon after each ErrorType

server/src/models/errors/api-error.ts
```diff
@@ -1,29 +1,42 @@
...
+    let msg: string = ERROR_MESSAGES[messageKey];
+    let argCounter = 0
+    for (let arg of args) {
+      msg = msg.replace(`{${argCounter}}`, arg);
+      argCounter ++;
+    }
+    return msg;
+  }
+
+  forResponse = (): Object => {
```
flashback2k14: nice catch. only this function declaration works. 👍  the old forResponse(): Object {..} function declaration throws not found errros!

server/src/logic/utils/error.utils.ts
```diff
@@ -7,4 +9,16 @@ export class ErrorUtils {
   static notFound (err: any): boolean {
     return err.code === errorCode.noData;
   }
+
+  static handleError(error: any, source: string, sourceMethod: string): ApiError{
```
flashback2k14: i like this error creation more than the long catch blocks. what did you think about use the builder pattern?

server/src/logic/services/auth.service.ts
hetec: Ok

server/src/errors.ts
hetec: Ok

server/src/errors.ts
hetec: Ok

server/src/error-messages.ts
hetec: Ok

server/src/logic/utils/error.utils.ts
hetec: Sounds good and  would End in a nice fluent style

hetec:
I hope that's better. It is not exactly the builder pattern because I think this would be overkill for the error object but I think the fluent style of adding properties is much better than setting each property separatly

flashback2k14:
Yes, this is much nicer. I like it.

hetec:
What do you think about the new way of creating errors? Is this near to your idea of the builder pattern or had you something different in mind?

flashback2k14:
I like it. We could merge this.

Issue #18, opened 2016-12-21
Add a new recipe


hetec:
- route for adding a new recipe
- create database service for recipes and implement the add function
- create db table

flashback2k14:
i think we should first implement user login and registration!?

hetec:
I think we can so this in parallel since user login should bei independent from the recipe service. Maybe a middleware which Hooksiel into the request chain before each recipe operation or so.

flashback2k14:
Yes, you are right. But before I would complete the error handling.

hetec:
Yes. You changes to the error  Handling look fine for me. I just have a question related to the spy. But im currently in the train anderen will ask you later via hangouts. If you are also happy with the state of the error handling you are encouraged to merge the PR 😉

PR #41, opened 2017-01-17
#18 - add new recipe

hetec:
Open problems see #42 

Add recipe service
- add save method for a recipe
- add get by id method (to simplify manual tests)
- add tests for the recipe service
Add recipe handler 
- Introduce the concept of a separate handler class instead of putting everything into the routes class
- Handler for save and getById
- Add tests for the handler
Add input validation
- Add validation middleware
- Add validation for recipes
- Add a validation handler
Move the whole error logic related to db operations into the db layer (includes error logic for the user service. I HOPE THATS OK?
Reason: I noticed that if we handle errors in a higher layer that we have to duplicate error logic if we want to use one service into another one. E.g I wanted to user the user service into the recipe service. If I wouldn't have moved the error logic I would had to write the error handling for getUserById in the recipe handler again. I Hope that makes sense.
- Add auth to the recipe
- Add a handler class for the auth validator

SOORRRY for this huge PR


hetec:
I think we could extract this later if we really need it. But you're right and we should keep this in mind

server/src/logic/services/user.service.ts
```diff
@@ -1,5 +1,6 @@
 import { IDatabase } from "pg-promise";
```
flashback2k14: Okay. New Error handling break the register logic.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+// Platzieren Sie Ihre Einstellungen in dieser Datei, um Standard- und Benutzereinstellungen zu überschreiben.
+{
+  "editor.tabSize": 2,
+  "editor.insertSpaces": false
```
hetec: Do you prefer more inline code in contrast to more LoC?

server/src/logic/middleware/validator.middleware.ts
```diff
@@ -0,0 +1,36 @@
+import { Router, Request, Response, NextFunction } from "express";
```
hetec: I think I found them \°/

server/src/logic/services/user.service.ts
hetec: Sorry I have missed this. I will fix it

server/package.json
```diff
@@ -30,29 +31,33 @@
...
     "@types/body-parser": "~0.0.33",
-    "@types/cors": "~0.0.33",
-    "@types/mocha": "~2.2.33",
     "@types/chai": "~3.4.34",
     "@types/chai-http": "~0.0.29",
+    "@types/cors": "~0.0.33",
+    "@types/express": "~4.0.34",
+    "@types/mocha": "~2.2.33",
     "@types/pg-promise": "~0.0.28",
-    "@types/winston": "~0.0.32",
```
hetec: I didn't want to remove this. Maybe this happened during the merge

server/src/logic/services/recipe.service.ts
```diff
@@ -0,0 +1,52 @@
...
+                           returning id`;
+
+  constructor (private _db: IDatabase<any>) {}
+
+  getById (id: number): Promise<Recipe> {
+    console.log("CAll getById");
+    let promise = this._db.one(this._GET_BY_ID, [id]);
+    return promise.catch(error => {
+      return ErrorUtils.handleDbNotFound(
+        ERRORS.RECIPE_NOT_FOUND,
```
hetec: Would you simply inline this?

flashback2k14: I would break the line if I hit the 80 rule or between 80 and 120.

hetec: So you would also keep multiple return statements inline? And should I alway use the promise pattern like this
promise()
  .then(...)
  .catch(...)
Like you mentioned below?

flashback2k14: Nope, the multiply returns are okay. I mean the function parameters. 

Hmm, that you only handle the catch has me confused a little bit. I haven't seen / used this before. But it works. It should be okay.

hetec: OK! I only handle the catch because I only need to manipulate the error. In the case of then the original Promise will be returned

hetec: I also had to read a bit about Promises to get this. But this works since each call to then or catch also returns a new Promise. And if you return sth from inside a Promise this gets the value of the newly created Promise. Its really a bit crazy!

server/test/validator.test.ts
```diff
@@ -0,0 +1,78 @@
...
+import * as chai from "chai"; 
+import * as sinon from "sinon";
+import {Router, Request, Response, NextFunction } from "express";
+
+import { ApiValidator } from "./../src/logic/middleware/index";
+import { Recipe, ValidationError } from "./../src/models/index";
+
+let mocks = require("node-mocks-http");
+require("mocha-as-promised")();
+var chaiAsPromised = require("chai-as-promised");
```
hetec: Oh yes this var is bad and the let could also be a const

server/src/logic/services/user.service.ts
flashback2k14: Sorry, for the lack of tests.

hetec:
Thanks for your suggestions and comments! I hope I caught everything.

server/src/logic/services/user.service.ts
flashback2k14: hmm, i think the register need same refactoring. this not working as before. it should return the errortype AUTH_USED_PASSWORD_EMAIL, but it returns Internal server error.

this is because you handle the user db errors inside the user.service and not simply return the promise.

flashback2k14: i have opened a new issue for this #44.

hetec: How do you get this error? If I try to register twice with the same email I got the error AUTH_USED_PASSWORD_EMAIL

Issue #38, opened 2017-01-15
add typescript linting


flashback2k14:
- [tslint Rules](https://palantir.github.io/tslint/rules/)
- [default Rules](https://github.com/palantir/tslint/tree/master/src/configs)

flashback2k14:
short reminder for myself:

<details>
  <summary>stashed changes</summary>
  stash@{0}: WIP on 12-user-management: 4779415 Merge pull request #32 from socobo/27- improve-error-handling
</details>

hetec:
I really agree with this issue. So i hopefully don't forget so many semicolons ;)

flashback2k14:
Hehe. Yes, linting is really helpful. :-)

PR #45, opened 2017-01-20
#38 - add typescript linting

flashback2k14:
ohh boi. that was really many linting errors.

@hetec could you please look at the test from RecipeService: 
_RecipeService save should open a transaction and call the one method of pgPromise with the correct query and parameters:_

this is failing, but i can't understand the test or the error. 😕


flashback2k14:
and we should sqash and merge this PR!

hetec:
I have fixed the test. You have added a name to the tx so the stub wanted to call the name as a function since the order of the arguments had changed. I just fixed the index number and the format of the sql string.

hetec:
Maybe you want to fix the to little things I had mentioned above. If not, feel free to merge this PR

Issue #33, opened 2017-01-13
Bug if you want to login with a username


hetec:
If several users were registered with email addresses like:

john@gmx.com,
john@gmail.com,
john@...

each of these accounts will get the username john. So if you now try to login as john an internal server error is thrown which tells you that multiple rows are not allowed. The query founds multiple rows because the username john exists 3 times in this case but should be unique.

"name": "QueryResultError",
      "stack": "QueryResultError: 2\n    at new QueryResultError (/Users/patrick/dev/socobo/SocoboPlusPlus/server/node_modules/pg-promise/lib/errors/queryResult.js:121:22)\n    at Query.<anonymous> (/Users/patrick/dev/socobo/SocoboPlusPlus/server/node_modules/pg-promise/lib/query.js:156:41)\n    at Query.handleReadyForQuery (/Users/patrick/dev/socobo/SocoboPlusPlus/server/node_modules/pg/lib/query.js:106:10)\n    at Connection.<anonymous> (/Users/patrick/dev/socobo/SocoboPlusPlus/server/node_modules/pg/lib/client.js:163:19)\n    at emitOne (events.js:101:20)\n    at Connection.emit (events.js:188:7)\n    at Socket.<anonymous> (/Users/patrick/dev/socobo/SocoboPlusPlus/server/node_modules/pg/lib/connection.js:109:12)\n    at emitOne (events.js:96:13)\n    at Socket.emit (events.js:188:7)\n    at readableAddChunk (_stream_readable.js:176:18)",
      "message": "Multiple rows were not expected.",

flashback2k14:
Hmm. Okay, than only register a new user is possible with E-Mail and provide an unique auto generated username like user123421. This username could be changed with the coming profile endpoint! What did you think @hetec?

hetec:
I'm with you that we should enable the user to set the username in the user profile. Maybe we could initially set the user name also to the email address, so it would be unique and the user can change this to what ever he want in the profile later.

What

flashback2k14:
Yes, thats even better, because we doesn't need extra logic to generate usernames.

hetec:
I think we could also do this after merging #12

flashback2k14:
Okay, but before merging #12 I would implement extend SocoboUser #28. Or should we merge #12 into master first?

hetec:
You can do this first because I think #28 is very related to #12

Issue #44, opened 2017-01-18
refactoring user registration


flashback2k14:
- [x]  use ExtractRequestBodyResult for authService.register and authService.login function parameter
- [x] only allow registration with email
- [x] register new user inside promise.then() not promise.catch() --> this could maybe fixed with issue #39

hetec:
How do you get this error? If I try to register twice with the same email I got the error AUTH_USED_PASSWORD_EMAIL

flashback2k14:
yes, you are right. the error come from the username. i have tested with a wrong email.

```json
{
    "name": "error",
    "source": "UserService",
    "sourceMethod": "save(..)",
    "error": {
      "name": "error",
      "length": 219,
      "severity": "ERROR",
      "code": "23505",
      "detail": "Key (username)=(JohnDoe) already exists.",
      "schema": "public",
      "table": "socobo_user",
      "constraint": "socobo_user_username_key",
      "file": "nbtinsert.c",
      "line": "424",
      "routine": "_bt_check_unique"
    },
    "message": "Internal server error",
    "statusCode": 500,
    "code": "00001",
    "timestamp": 1484776451495
  }
```

flashback2k14:
but i think the registration inside the catch is not so nice!?

PR #46, opened 2017-01-22
#33 - bug if you want to login with a username

flashback2k14:
- change registration behavior
  - if registration w/ email, then username is equal to email and Provider is set to Email
  - if registration w/ username, then email is equal to username and Provider is set to Username


hetec:
Works fine for me! I just added a little fix. But there is still a linter error and I don't know how to fix since it looks fine for me.
test/users.route.test.ts[6, 10]: Named imports must be alphabetized

flashback2k14:
p comes before s ;-)

hetec:
In general I'm still not completely happy with the mechanism of registration. I would prefer a registration where we always require the email address and a password. The username can then be provided additional to the email by the user. If the user provides a name we set the the name if not we use the email as name.

I think it's a bit odd to use the username as email since we cannot send emails to a username and we cannot be sure that we get a email from the user in the future.

What do you think? (Would be the subject of another issue)

hetec:
Ups I looked in the wrong file :)

lyio:
Without having been involved in anything so far, I agree with @hetec. We should always require an email address as that 
1. is our only way to contact the user
2. is the users way to reset their password

flashback2k14:
yes you are right. we should change this.

hetec:
Maybe later, but I can so it

flashback2k14:
Okay, but the bug is fixed. Maybe we should change the behavior with issue #44!?

Issue #42, opened 2017-01-17
Remove user from the request body


hetec:
We have a problem with the auth validator. The user is directly appended to the req.body. But if we want to add resources like recipes or other item the user is mixed into these object and we get something like this:

{
  "title": "test",
  "userId": 1,
  "description": "Test Desc",
  "imageUrl": "http://someimage.jpg",
  "decoded": {
    "id": 3,
    "username": "pat",
    "email": "pat@test.com",
    "password": "$2a$10$N7vJc1VUgbrq7nEgmCMPXugRg.YC6RviQ6ZWQETZW2F1edGnWPkmu",
    "image": "http://placehold.it/350x150",
    "hastermsaccepted": true,
    "isadmin": true,
    "provider": "email",
    "created": "1484548879642",
    "lastmodified": "1484548879642",
    "iat": 1484599012,
    "exp": 1484685412
  },
  "created": "2017-01-17T20:11:06.841Z",
  "id": 24
}

Decoded should not be part of the recipe! I think we should add the user to the req object itself rather than to the body object?!

flashback2k14:
Yes, this is not good. I have added this to the body for type compatibility. If we add decoded to req, than we must use req: any instead of req: Request as function parameter. But this should be fixed if issue #34 is implemented. I have planned to remove the decoded object after the user check.

hetec:
Ok if #34 will fix this everything is fine

hetec:
I will close this issue since this was fixed in #34

Issue #34, opened 2017-01-14
improve JWT creation and validation


flashback2k14:
- create JWT only with specific user fields
- create new middleware function inside AuthValidator called checkPayload() 
  - if the extracted user is a valid user 
  - if the user is allowed to access a specific endpoint, e.g. admin

hetec:
Sounds good! What fields should go into the jwt? I saw that I had forgot to change the error in the auth validator. Maybe you could adapt this in the scope of this issue too.

flashback2k14:
i would follow the [JWT claims](https://tools.ietf.org/html/rfc7519#section-4.1). 
```json 
{
  "iss": "socobo",
  "exp": "comes from the jwt module",
  "iat": "comes from the jwt module",
  "username": "user1234",
  "email": "mail@test.test",
  "admin": "true|false"
}
```

what did you think? is this enough? @hetec 

yes, i change the error object creation in this issue.

hetec:
I think this is ok! I think we should add more public claims first if we need them.

flashback2k14:
Yes, this is enough for this time.

PR #47, opened 2017-01-22
#34 - improve jwt creation and validation

flashback2k14:
much bigger PR as expected

- auth middleware
  - check valid token
  - check valid user
  - restrict routes --> now error logs route only available for admin users
- general changes
  - fix inconsistent class and file names
  - add Handler classes for Auth, Users and Logs
  - reduce JWT Payload


flashback2k14:
What did you think about the changes @hetec?

server/src/logic/handler/validation/auth-validation.handler.ts
```diff
@@ -0,0 +1,44 @@
...
+import { ApiError, DbError, Recipe, ValidationError } from "./../../../models/index";
+
+export class AuthValidationHandler {
+
+  constructor (private _authValidationMiddleware: AuthValidationMiddleware) {}
+
+  public checkRequest = (req: Request, res: Response, next: NextFunction): void => {
+    this._authValidationMiddleware.checkRequest(req)
+      .then(() => next())
+      .catch((e: any) => res.status(e.statusCode).json(e.forResponse()));
```
hetec: since this seems to be always the same you could give this function a name and just use this name in the catch blocks

hetec: maybe const sendErrorResponse = (e:any) => {...}

server/src/logic/middleware/auth-validaton.middleware.ts
```diff
@@ -0,0 +1,129 @@
...
+import * as jwt from "jsonwebtoken";
+import { Config } from "./../../config";
+import { ApiError, ERRORS, ExtractRequestBodyResult, SocoboUser } from "./../../models/index";
+import { UserService } from "./../services/index";
+
+export class AuthValidationMiddleware {
+
+  constructor (private _userService: UserService) {}
+
+  public checkRequest (req: Request): Promise<any> {
```
hetec: I still think we should do this with the modelValidation. Maybe we need something like a credentials model and add the constraints to the attributes of this model (username, email, password or so). I think so because you still accept all kinds of emails (also wrong ones with out @ and so) and the password and username would also need some additional validation like min length or allowed characters. And I don't think that you want to implement this by yourself ;)

What do you think. Maybe we should adapt this later.

server/src/logic/middleware/auth-validaton.middleware.ts
```diff
@@ -0,0 +1,129 @@
...
+        .then((user: any) => {
+          delete req.body.decoded;
+          req.body.isAdmin = user.isadmin;
+          resolve();
+        })
+        .catch((error: any) => reject(error));
+    });
+  }
+
+  public checkUserRoleForRestriction (req: Request, shouldBeRestricted: boolean): Promise<any> {
```
hetec: Is this really necessary? You already check in the the checkValidUser for the isAdmin property. So you could simply check if the user is admin in this method and send a error response directly from this method if not.
This means, if you want to restrict a endpoint you would simply add the checkValidUser (which does the check) and if you don't want to restrict the endpoint you could simply leave the checkValidUser() out.

Just an idea

server/src/logic/middleware/auth-validaton.middleware.ts
flashback2k14: You are right. I didn't implement this by myself. I would use the modelValidationMiddleware and add the needed decorators inside SocoboUser. class-validatior has a decorator for email.

This should be located into another issue.

server/src/logic/handler/routes/auth.handler.ts
```diff
@@ -0,0 +1,50 @@
...
+  public login = (req: Request, res: Response): void => {
+
+    if (!req.body.hasOwnProperty("ExtractRequestBodyResult")) {
+      const err: ApiError = new ApiError(ERRORS.REQUEST_BODY)
+        .addSource(AuthHandler.name)
+        .addSourceMethod("login");
+      res.status(err.statusCode).json(err.forResponse());
+    }
+
+    const erbr: ExtractRequestBodyResult = req.body.ExtractRequestBodyResult;
```
flashback2k14: You are right. I will create a function for this.

server/src/logic/handler/validation/auth-validation.handler.ts
flashback2k14: I will add this.

server/src/logic/middleware/auth-validaton.middleware.ts
flashback2k14: Hmm, the checkValidUser() is for checking if the user is inside our database and remove the decoded object after a successfully result. This close issue #42. and this function is not optional. 

How would you check inside this chain if the user is an admin or whatever user roles we are adding in the future?

hetec: I will build this in a separate branch to show you what I mean.

I currently noticed that isAdmin is alway set to true. Is this a bug?

flashback2k14: okay. 

yes, i change this in a few minutes.

hetec:
Your changes were fine and I think we can also merge this PR, if all checks are passed

PR #51, opened 2017-02-01
#34 refactoring idea

hetec:
Just a little hack to demonstrate the idea:

- remove the checkUserRoleForRestriction method
- use only checkValidUser method to authorize a user
How it works:
- Fetch the user from the db and get his isadmin property
- check if is admin is true
- If you add this middleware to the route it will be checked if the user is a admin
- If you leave out the middleware no check will be done

Roles:
- different roles can be provided via a parameter of the checkValidUser middleware
- so you can add the role which would be required for the current resource
- We could provide a enum with existing roles or fetch them from the db or so


hetec:
This is just a hack, so I didn't have removed comments and so on. Refactoring would be necessary.

server/src/routes/api/v1/recipes/recipe.route.ts
```diff
@@ -17,14 +17,14 @@ export class RecipeRoute {
...
       this._authValidationHandler.checkToken,
       this._authValidationHandler.checkUser,
-      this._authValidationHandler.nonRestricted,
+      //this._authValidationHandler.nonRestricted,
       this._modelValidationHandler.validate,
       this._recipeHandler.save);
 
     this._router.get("/:id",
       this._authValidationHandler.checkToken,
-      this._authValidationHandler.checkUser,
```
flashback2k14: If you do this, than is the decoded object still inside the request body. This is the reason why I have added the che kUser middleware.

hetec: Sometimes yes, but not in this case since this only happen with the post method.
(I will try to find a solution that works always)

Do we need to check the user in each case? If you want to check the user you could simply add the checkUser with the role 'user' or so.

hetec: I have added an approach to make the request itself extendable, so we don't have to remove the properties and they don't pollute the body

hetec: What do you think about this? This seems to be a valid approach since it is suggested in a lot of stackoverflow answers

flashback2k14: Looks really good! +1

server/src/logic/services/auth.service.ts
```diff
@@ -137,9 +137,11 @@ export class AuthService {
         .addPassword(hashedPassword)
         .addImage((process.env.DEFAULT_USER_IMAGE || Config.DEFAULT_USER_IMAGE))
         .addHasTermsAccepted(true)
-        .addIsAdmin(true)
+        .addIsAdmin(isAdmin)
```
flashback2k14: I would add an enum "UserRole" and change addIsAdmin to addUserRole or simply role!?

flashback2k14:
What did you think about my changes @hetec ?

hetec:
I will check your changens soon! I promise! Currently I'm just a bit busy, sorry!

Issue #39, opened 2017-01-15
DB layer refactoring


hetec:
See also #37

flashback2k14:
I think we should start the implementation after the issues #33, #34, #37 and #42.

flashback2k14:
- [ ] #33 
- [ ] #34 
- [ ] #38
- [ ] #42

flashback2k14:
@hetec After closing all referenced issues we can start the refactoring.

hetec:
Yes I also think that we're ready now. Do you want to do this or should I do it?

flashback2k14:
You can start the refactoring and I look into your commits and tell you what I like or dislike. ;-)

hetec:
Sound good!

PR #52, opened 2017-02-17
#39 db layer refactoring

hetec:
So I hope I have implemented the idea behind the db refactoring right ;)

Some not covered points:
- Currently only inline sql will be used
- I'm missing the IOption type for the db setup, so I currently use any
- There are still a few linting errors I didn't got fixed - help wanted!


flashback2k14:
I have a look at it and fix the errors.

server/src/config.ts
```diff
@@ -1,10 +1,29 @@
-/* tslint:disable:max-line-length */
+import { DbConfig } from "./models/index";
+
 export class Config {
   public static DB_USER: string = "postgres";
   public static DB_PW: string = "socobo";
-  public static DB_URL_TEST: string = `postgres://${Config.DB_USER}:${Config.DB_PW}@localhost:5432/socobo_test?ssl=false`;
-  public static DB_URL_DEV: string = `postgres://${Config.DB_USER}:${Config.DB_PW}@localhost:5432/socobo_dev?ssl=false`;
-  public static DB_URL: string = `postgres://${Config.DB_USER}:${Config.DB_PW}@localhost:5432/socobo?ssl=false`;
+  public static DB_CONFIG_TEST: DbConfig = {
```
flashback2k14: i don't know if it possible to set an object as environment variable!?

server/src/db/repositories/recipe.repository.ts
```diff
@@ -0,0 +1,47 @@
...
+                                from recipes
+                                where recipes.id = $1`;
+
+  private _SAVE: string = `insert into recipes(
+                             title, userId, description,
+                             imageUrl, created)
+                           values($1, $2, $3, $4, $5)
+                           returning id`;
+
+  constructor (private _db: any) {}
```
flashback2k14: after this change we have no intellisense for the database methods, but we can't do this:

```typescript
constructor (private _db: IDatabase<any>) {}
```

we need to do this:

```typescript
private _db: IDatabase<any>;

constructor (db: any) {
  this._db = db;
}
```

Should we change this?

server/src/db/repositories/user.repository.ts
```diff
@@ -0,0 +1,91 @@
+import { IDatabase } from "pg-promise";
+import { ErrorUtils } from "./../../logic/utils/index";
+import { DbError, ERRORS, Recipe, SocoboUser } from "./../../models/index";
+
+export class UserRepository {
+
+  constructor (private _db: any) {}
```
flashback2k14: same as RecipeRepository

server/src/logic/handler/routes/recipe.handler.ts
```diff
@@ -1,16 +1,14 @@
 import { Request, Response } from "express";
-import { RecipeService, UserService } from "./../../../logic/services/index";
 import { Recipe } from "./../../../models/index";
 
 export class RecipeHandler {
 
   constructor (
-    private _recipeService: RecipeService,
-    private _userService: UserService
+    private _db: any
```
flashback2k14: there is no intellisense for this._db.recipes or this._db. :-(

server/src/logic/handler/routes/user.handler.ts
```diff
@@ -1,19 +1,20 @@
 import { Request, Response } from "express";
-import { UserService } from "./../../../logic/services/index";
+import * as db from "./../../../db/index";
 import { SocoboUser } from "./../../../models/index";
 
 export class UserHandler {
 
-  constructor (private _userService: UserService) {}
+  constructor (private _db: any) {}
```
flashback2k14: same as RecipeHandler

server/src/logic/middleware/auth-validaton.middleware.ts
```diff
@@ -2,11 +2,10 @@ import { NextFunction, Response } from "express";
 import * as jwt from "jsonwebtoken";
 import { Config } from "./../../config";
 import { ApiError, ERRORS, ExtractRequestBodyResult, Role, SocoboRequest, SocoboUser } from "./../../models/index";
-import { UserService } from "./../services/index";
 
 export class AuthValidationMiddleware {
 
-  constructor (private _userService: UserService) {}
+  constructor (private _db: any) {}
```
flashback2k14: same as the other handlers

server/src/db/repositories/recipe.repository.ts
hetec: I have implemented this as in the example of pg-promise to get a first version running. But I also think we should introduce this type. I will change this. Maybe we should use the exact same type as the exported db value: <IDatabase<IExtensions>&IExtensions> if this will work?!

flashback2k14: my example is also from the demo repo, but i have reordered the private variable. Yes, we should change the any to the right object.

server/src/config.ts
hetec: Ok yes, I forgot that we use some of the values as env variables. I just wanted to access the db config object from inside the index.ts where the db is set up. Do we really need to set DB_CONFIG as env variable?

flashback2k14: Hmm, I think yes. Otherwise we commit our real db connectionstring to github!? what if we implement a toString() function?

server/src/db/repositories/recipe.repository.ts
hetec: Oh Oh, I never had seen these variable declarations since they are placed under the constructor. So I thought only the any type was used. That happens if I try to do this as fast as possible in the evening. Sorry.

server/src/config.ts
flashback2k14: Okay the toString() function doesn't help us there.

hetec: We already do this because we commit the Config file which contains the strings, or not? But you are right, that is maybe not the best idea!

flashback2k14: No, we commit our fake production connectionstring. At heroku or whatever hosting we are using we provide the right production connectionstring as env variable.

server/src/db/repositories/recipe.repository.ts
flashback2k14: No Problem. :-)

server/src/config.ts
hetec: Ah ok, now I understand your point and how you planned this! Ok in this case I would think we either need to stick with the url strings or we could also provide each value of the db config as a single env variable. Another approach would be to not use environment variables. Instead we could use a separate config file for the production environment which would not be committed to github.

So I think the url approach is not the most beautiful but the simples one and I would stick with this. What do you think? But the config file approach would be more flexible since we could change values at runtime without redeployment. What do you think?

flashback2k14: Hmmm, i think we should use the url string, because we doesn't need any logic changes or additions. The connectionstring is also flexible. I think we have enough infrastructure stuff and we should focus on implementing the real logic!? ;-)

hetec: Yes I'm with you. I just wanted to point out the opportunities ;) I will change this back to the urls

flashback2k14: hehe. too much opportunities. :-D 👍

flashback2k14:
yippie. intellisense is back! 👍

flashback2k14:
If you have no additions than I would say that we could merge this!?

PR #53, opened 2017-02-18
#39 - db layer refactoring addition

flashback2k14:
i have the following changed:
- move interfaces to models/generics folder --> maybe create an interfaces folder
- bump pg-promise version to get access to IOptions
- remove @types/pg-promise, because pg-promise has it's own typings
- again workarounded the SocoboUser Object creation problem with ObjectUtils
  - i'm totally unhappy with this, because pg-promise returns always a plain object for SocoboUser and Recipe with a different structure then our typescript objects (all properies are lowercase and all enum values are strings)
  - the ObjectUtils doesn't really fix that problem :-(
  - i have no clue how to fix this!
- Rename Recipe.Service.Test to Recipe.Repository.Test
  - Recipe Repository Test throws some warnings:
<details>
<summary>Warnings</summary>
WARNING: Creating a duplicate database object for the same connection.
    at Context.<anonymous> (/Users/flbk/Documents/Development/Web/socobo/SocoboPlusPlus/server/test/recipe.repository.test.ts:22:10)
    at callFn (/Users/flbk/Documents/Development/Web/socobo/SocoboPlusPlus/server/node_modules/mocha/lib/runnable.js:345:21)
    at Hook.Runnable.run (/Users/flbk/Documents/Development/Web/socobo/SocoboPlusPlus/server/node_modules/mocha/lib/runnable.js:337:7)
    at next (/Users/flbk/Documents/Development/Web/socobo/SocoboPlusPlus/server/node_modules/mocha/lib/runner.js:309:10)
    at Immediate.<anonymous> (/Users/flbk/Documents/Development/Web/socobo/SocoboPlusPlus/server/node_modules/mocha/lib/runner.js:339:5)
</details>


hetec:
Thanks for the refactoring!

I don't really understand how pg-promise can assign a object with no type and different properties to our SocoboUser since the Promise returned by our user repository defines the Promise to have a Object with type Socobo User in it?! That's really annoying!

Could you elaborate on what the concrete problems is if you want to use the returned SocoboUser? Is it the fact that you can't use the properties as they are defined or do you get some errors if you try to use the returned object from pg-promise?

flashback2k14:
Okay. If we wright our code with user: SocoboUser then we can use all the great stuff that vs code provide us (autocompletion, etc.) but if we run the code and debug the code than you will see that user: SocoboUser does nothing. user is always a Object and has only lowercase properties like hastermsaccepted = "true" as expected result should be hasTermsAcceted = true. 

You can see this inside the auth.service _validateUser function.

hetec:
Yes I have already seen this. So the main problem is that debugging does not work correctly without types?
Recently I also tried to use the debugger but at some points it was not able to set breakpoints correctly or stop the code execution. Maybe this happened also due to the missing types.

Ok, but I still don't know how this can happen. To my understanding typescript should complain about the wrong type since the lowercase properties don't match our SocoboUser type. Hmm.

Another problem is that the properties are also displayed wrong in the response.

flashback2k14:
Yes, the whole typecasting doesn't work as expected. It's maybe a problem inside our tsconfig.json. 

The properties are displayed wrong, because the pg-promise output.

hetec:
Maybe. The best way to solve this would be if we could tell the pg-promise query which type to return explicitly. But unfortunately there seems not to be such an option.

Maybe we should live with your fix for the moment

flashback2k14:
Yes, i try a better solution tomorrow. Please don't merge this at the moment.

flashback2k14:
@hetec now i'm happier with the solution. our repositories now returning the expected typescript objects. 👍

vitaly-t:
> The best way to solve this would be if we could tell the pg-promise query which type to return explicitly. But unfortunately there seems not to be such an option.

Actually, there is. All database methods allow explicit type override, see optional parameters `cb` and `thisArgs`, also see methods `map` and `each`.

http://vitaly-t.github.io/pg-promise/Database.html

flashback2k14:
thanks @vitaly-t  for your help. After reading this [tutorial](http://cs.mcgill.ca/~mxia3/2016/11/18/Statically-typed-PostgreSQL-queries-and-typescript-schemats/) if have seen that we could add a callback function to transform the result into your typescript objects. 👍

hetec:
Yes this solution is much better! But it is unfortunately still a lot work to do, to only transform the types. I hoped there would be a solution where I can simply pass an interface or the like to the db method and pg-promise would care about the transformation. But this solution works fine and is maybe more flexible.

flashback2k14:
>I hoped there would be a solution where I can simply pass an interface or the like to the db method and pg-promise would care about the transformation. 

I think this is not possible, because under the hood is JavaScript logic working and JavaScript didn't know what an interface is!? @vitaly-t

vitaly-t:
>But it is unfortunately still a lot work to do, to only transform the types. I hoped there would be a solution where I can simply pass an interface or the like to the db method and pg-promise would care about the transformation. But this solution works fine and is maybe more flexible.

Check out [schemats](https://github.com/SweetIQ/schemats), it might be what you are looking for ;)

flashback2k14:
@hetec i think at the moment we have no need to introduce another module!?

vitaly-t:
> i think at the moment we have no need to introduce another module!?

Node.js is in a way a midset to embrace high modulation. The 10m+modules on NPM will attest to that :wink:

hetec:
Yes, we should stick with the current state for the moment. Maybe we can introduce this if our number of repositories grows. So we should keep this in mind.

I think we can merge this PR now

flashback2k14:
Yes. We could merge this.

vitaly-t:
Were you guys actually using the patters advocated in [pg-promise-demo](https://github.com/vitaly-t/pg-promise-demo)?

flashback2k14:
@vitaly-t Please look at PR #52.  We have started using this pattern. Maybe you have some suggestion or additions for us!?

vitaly-t:
No, I was just looking that your are not using that external SQL approach ;)

PR #56, opened 2017-02-22
#44 - refactoring user registration

flashback2k14:
i haven't found any good solution for "register new user inside promise.then() not promise.catch()".
i think this not really needed!?


flashback2k14:
@hetec short reminder :-)

hetec:
I try to check your work as soon as possible.

PR #58, opened 2017-03-08
57 additional recipe crud operations

hetec:
Add new endpoints or extend existing ones:

- delete by id
- get all recipes
- get all filtered by property (this includes get by userId)
- search for a property values (search works only inside one property, a general search comes later)
- update (also particular properties)

Extend the recipe with foodItem and recipeStep (new models)
- These models are not directly set via the recipe class
- I think we need to create separate repositories for them (new issues)
- And I'm not sure if we maybe want to create separate endpoints for them?! So we could load them independently from the recipe


server/README.md
```diff
@@ -286,6 +286,123 @@ Error body:
...
+  ```
+
+**DELETE /api/v1/recipes/:id**
+
+Deletes the recipe found for the given id
+
+Path Parameter:
+  - id: Recipe id
+  
+  ```
```
flashback2k14: this breaks the readme formatting.

server/src/logic/handler/routes/recipe.handler.ts
```diff
@@ -17,6 +20,60 @@ export class RecipeHandler {
       .catch((e: any) => res.status(e.statusCode).json(e.forResponse()));
   }
 
+  public getAll = (req: Request, res: Response): void => {
+
+    const queryPramas = req.query;
+    const allQueryParams = Object.getOwnPropertyNames(queryPramas);
+    const firstQeryParam = allQueryParams[0];
```
flashback2k14: small typo :-)

server/src/logic/handler/routes/recipe.handler.ts
```diff
@@ -17,6 +20,60 @@ export class RecipeHandler {
...
+        res.status(e.statusCode).json(e.forResponse());
+      }
+      this._db.recipes.getByField(firstQeryParam, valueFirstQueryParam)
+      .then((result: Recipe[]) => res.status(200).json(result))
+      .catch((e: any) => res.status(e.statusCode).json(e.forResponse()));
+    } else {
+      this._db.recipes.getAll()
+      .then((result: Recipe[]) => res.status(200).json(result))
+      .catch((e: any) => {
+        if (e.statusCode === 404) {
```
flashback2k14: hmm, i'm not sure if i like this.

server/src/logic/handler/routes/recipe.handler.ts
```diff
@@ -31,4 +88,45 @@ export class RecipeHandler {
...
+
+  public update = (req: Request, res: Response): void => {
+    const newRecipe: Recipe = req.body as Recipe;
+
+    this._db.users.getUserById(newRecipe.userId)
+      .catch((e: any) => res.status(e.statusCode).json(e.forResponse()));
+
+    this._db.recipes.update(req.params.id, newRecipe)
+      .then(() => {
+        this.getById(req, res);
```
flashback2k14: this could be a one liner!?

server/src/logic/handler/routes/recipe.handler.ts
```diff
@@ -17,6 +20,60 @@ export class RecipeHandler {
...
+
+    if (firstQeryParam && valueFirstQueryParam) {
+      if (!(new Recipe().fields.has(firstQeryParam))) {
+        const e = new ApiError(ERRORS.RECIPE_INVALID_FIELD.withArgs(firstQeryParam))
+          .addSource(RecipeHandler.name)
+          .addSourceMethod("getAll()");
+        res.status(e.statusCode).json(e.forResponse());
+      }
+      this._db.recipes.getByField(firstQeryParam, valueFirstQueryParam)
+      .then((result: Recipe[]) => res.status(200).json(result))
```
flashback2k14: one tab is missing!

server/src/logic/handler/routes/recipe.handler.ts
```diff
@@ -31,4 +88,45 @@ export class RecipeHandler {
...
+      .catch((e: any) => res.status(e.statusCode).json(e.forResponse()));
+
+    this._db.recipes.update(req.params.id, newRecipe)
+      .then(() => {
+        this.getById(req, res);
+      })
+      .catch((e: any) => res.status(e.statusCode).json(e.forResponse()));
+  }
+
+  public updateRecipe = (req: Request, res: Response, next: NextFunction) => {
```
flashback2k14: i think this should be moved into a middleware. maybe into the ModelValidationMiddleware class.

hetec: This was also my first Idea but the function is only used in recipes and does not really some validation stuff. I kept it in the handler because I didn't want to create a separate middleware for one single function related to one special part of the app (recipe). What do you think, is it worth to create this new class for one function?

server/src/logic/handler/routes/recipe.handler.ts
hetec: I will compare the error code (from the ERRORS class), I think this will be more specific.

flashback2k14: Okay.

server/src/logic/handler/routes/recipe.handler.ts
flashback2k14: I think yes. What does the modelutils function? If I get it right than Object.assign(..) does the same!?

npm-debug.log
```diff
@@ -0,0 +1,22 @@
+0 info it worked if it ends with ok
```
flashback2k14: please remove this.

server/src/logic/utils/model.utils.ts
```diff
@@ -0,0 +1,19 @@
+import { ApiError, ERRORS } from "./../../models/index";
+
+export class ModelUtils {
```
flashback2k14: What does the modelutils Class? If I get it right than Object.assign(..) does the same!?

flashback2k14: this could be simplified with Object.assign(..):

bildschirmfoto 2017-03-10 um 18 48 35

hetec: The model.utils takes a existing object (like a recipe already stored in the db) and a new value  (obj which contains properties which should be updated in the existing value). Then it checks if the provided properties are available and updatable in the existing object (the existing obj has a map with it's updatable fields). If the fields are available and updatable, these fields will be overwritten with the new values. The updated object will be returned. Therefor I cannot use Object.assign since this would also add new properties to the existing object if they do not exist already.

flashback2k14: Hmm, okay. But what if you change modelutils to modelmiddleware? :-) we have all middleware functions inside middleware classes.

hetec: I thought this is not really a middleware function because middleware functions typically take a res or / and a req object and are a part of the request chain.

This function should only be a simple helper. What's the definition of a middleware function?

flashback2k14: Sorry, I mean the this._recipeHandler.updateRecipe function.

hetec: Ok, you are right! It's maybe cleaner to also add this function to it's own middleware class

hetec:
I have extracted a middleware class for the update stuff. But I left the little handler in the recipe handler class. I hope that's ok

flashback2k14:
yes, i think it's okay at the moment.

PR #60, opened 2017-03-12
#50 - add validation to socobo user

flashback2k14:
- add validation decorators to SocoboUser
- introduce ValidationGroups - needed to use different validation scopes
- ModelValidationHandler.validate is now a generic function
  - needed Parameter
    - Type (Model Object)
    - ValidationGroup[] (Validation Scopes)
- ModelValidationMiddleware.validate needs now a new Parameter
  - ValidationGroup[] (Validation Scopes)
- ModelValidationMiddleware.validate changed behavior
  - then calls next()
  - catch throws Errors

### DIFF
--- a/server/src/models/database/recipe.ts
+++ b/server/src/models/database/recipe.ts
@@ -1,12 +1,13 @@
 import { IsNotEmpty, IsNumber, Length } from "class-validator";
+import { ValidationGroup } from "./../enums/validation-group";
 import { FoodItem, RecipeStep, SocoboUser } from "./../index";
-import { ValidationGroup } from "./../enums/validation-group"; 
 
 export class Recipe {
 
   public fields: Map<string, Function>;
 
   public id: number;
+
   @IsNotEmpty({
     groups: [ ValidationGroup.RECIPE ]
   })
@@ -14,6 +15,7 @@ export class Recipe {
     groups: [ ValidationGroup.RECIPE ]
   })
   public title: string;
+
   @IsNumber({
     groups: [ ValidationGroup.RECIPE ]
   })
--- a/server/src/models/database/recipe.ts
+++ b/server/src/models/database/recipe.ts
@@ -1,15 +1,21 @@
 import { IsNotEmpty, IsNumber, Length } from "class-validator";
-import { FoodItem, RecipeStep, SocoboUser } from "./../index";
+import { FoodItem, RecipeStep, SocoboUser, ValidationGroup } from "./../index";
 
 export class Recipe {
 
   public fields: Map<string, Function>;
 
   public id: number;
-  @IsNotEmpty()
-  @Length(1, 50)
+  @IsNotEmpty({
+    groups: [ ValidationGroup.RECIPE ]
+  })
+  @Length(1, 50, {
+    groups: [ ValidationGroup.RECIPE ]
+  })
   public title: string;
-  @IsNumber()
+  @IsNumber({
+    groups: [ ValidationGroup.RECIPE ]
+  })
   public userId: number;
   public description: string;
   public imageUrl: string;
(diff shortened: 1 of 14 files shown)


server/src/models/database/socobo-user.ts
```diff
@@ -1,16 +1,64 @@
...
+  })
+  @IsNotEmpty({
+    groups: [ ValidationGroup.LOGIN ]
+  })
+  @MinLength(5, {
+    groups: [ ValidationGroup.LOGIN ]
+  })
   public username: string;
+
+  @ValidateIf((o) => o.username === "", {
```
hetec: If a have a valid username and a invalid email the email will not be checked but is still used for authentication -> 401 (but I think it should be a 400)
Example body:
{
	"username": "test1@test.com",
	"email": "test1",
	"password": "testtest"
}

response:
"The user with the username/email test1 is not registered yet"

Maybe it would be better to check the values alway if they are present

server/src/models/database/socobo-user.ts
```diff
@@ -1,16 +1,64 @@
...
   public email: string;
+
+  @IsNotEmpty({
+    groups: [
+      ValidationGroup.LOGIN,
+      ValidationGroup.REGISTRATION
+    ]
+  })
+  @MinLength(8, {
+    groups: [
```
hetec: If I have something like this:
{
	"username": "",
	"email": "",
	"password": "testtest"
}
I got 
email must be an email

In some way this is correct, but is it valid that the email is empty during the login if I provide a email property? I think if there is a email property in the body it should not be empty as with the username

server/src/models/database/recipe.ts
```diff
@@ -1,15 +1,24 @@
...
 export class Recipe {
 
   public fields: Map<string, Function>;
 
   public id: number;
-  @IsNotEmpty()
-  @Length(1, 50)
+
+  @IsNotEmpty({
+    groups: [ ValidationGroup.RECIPE ]
```
hetec: Is it possible to simply use non validation group if there is only one possible as with the recipe (I cannot really think of another use case for this model) or do we need to always add a validation group?

server/src/models/database/socobo-user.ts
flashback2k14: This is not a valid body. What route have you called?

hetec: login

server/src/models/database/socobo-user.ts
flashback2k14: This is also not an valid request body.

server/src/models/database/socobo-user.ts
flashback2k14: Okay valid body are username and password or email and password.

server/src/models/database/recipe.ts
flashback2k14: Yes. I think its possible to not use a group. But I need to check the documentation first.

server/src/models/database/socobo-user.ts
hetec: Okay we can simply define this in the read me, but I thought this could maybe also be restricted directly by our api. Is there a easy way in to say that it is only valid to use one of the two? If not I think the current solution would be good.

flashback2k14: Please take a look into the readme. Its already there. Maybe we need to add an extra parameter to identify what login is called or a middleware!?

hetec: This should not sound like we should add this to the read me, I already have seen it there. I just wanted to say that it is fine to state this in the ready but we should maybe restrict it programmatically too ;)

flashback2k14: Hmm, maybe add a middleware to check the body for login requests. If I get it right than then decorators already restrict this!? Maybe the error message is not the best.

hetec: Okay, so I would not change any logic, I think this should be fine. A clearer error message would be nice but if this is not easily to accomplish, I would stick with the current version. Maybe this wouldn't be a problem at all since we are the only client at the moment and it is also in the readme. Sorry for the long discussion, I was just a bit confused by this. Maybe you can find a way to improve the message. If not lets go ahead and merge this PR.

flashback2k14: The decorators has also a message property for error messages. But I think let us merge this and improve this later.

PR #74, opened 2017-06-17
#70 - implement image upload

hetec:
- Create image service to store the image on the file system
- Image service is designed to be reused in for other endpoints (i.e. profile) and different technologies (we can add a implementation to store images at some cloud storage for instance)
- Create image upload endpoint, handler and middleware
- Update the image url in the related recipe database entity


server/README.md
```diff
@@ -424,14 +424,45 @@ Error body:
   }
   ```
 
+**POST /api/v1/recipes/:id/image**
+
+Uploads a provided image file to the server for the current user and updates
+the related recipe in the database.
+
```
flashback2k14: please provide an request body / header part!

server/package.json
```diff
@@ -36,34 +36,38 @@
     "cors": "~2.8.1",
     "es6-shim": "~0.35.2",
     "express": "~4.14.0",
+    "multer": "^1.3.0",
```
flashback2k14: please use ~ for your packages.

server/src/config.ts
```diff
@@ -14,5 +14,7 @@ export class Config {
   public static TOKEN_ISSUER: string = "socobo";
   public static TOKEN_HEADER: string = "x-access-token";
   public static DEFAULT_USER_IMAGE: string = "http://placehold.it/350x150";
+  public static IMAGE_TMP_DIR: string = "/Users/patrick/tmp";
```
flashback2k14: please remove /Users/patrick/ and look into server.ts comment. :-)

server/src/server.ts
```diff
@@ -179,11 +186,26 @@ class Server {
...
     this._logHandler = new LogHandler();
   }
 
   /**
+   * UPLOADS
+   */
+  private _uploader (): void {
+    const storage = multer.diskStorage({
+      destination: (req, file, cb) => {
+        cb(null, Config.IMAGE_TMP_DIR);
```
flashback2k14: i would change this to:
```ts
`${process.cwd()}/${process.env.IMAGE_TMP_DIR || Config.IMAGE_TMP_DIR}`)
```
this creates a new folder into our server folder or looks into it. i think this is better!?

server/src/logic/services/filesystem-image.service.ts
```diff
@@ -0,0 +1,56 @@
...
+      const userDataDir = `${Config.DATA_BASE_DIR}/${userIdentifier}`;
+      const dataTypeDir = `${userDataDir}/${dataType}`;
+      const targetPath = `${dataTypeDir}/${fileName}`;
+
+      const error = new ApiError(ERRORS.IMAGE_UPLOAD)
+        .addSource(FilesystemImageService.name)
+        .addSourceMethod("persistImage");
+
+      // 'this' doesn't work correctly inside of fs.readFile.
+      function createDirIfNotExists (dir: string, rej: Function): void {
```
flashback2k14: why have you added the function declaration into the promise? is this a common practice?

server/src/logic/handler/routes/recipe.handler.ts
```diff
@@ -115,4 +120,20 @@ export class RecipeHandler {
       .then(() => res.status(200).json())
       .catch((e: any) => res.status(e.statusCode).json(e.forResponse()));
   }
+
+  public uploadImage = (req: SocoboRequest, res: Response) => {
```
flashback2k14: i think this is looks really complicated. what did you think about this?

```ts
  public uploadImageV2 = (req: SocoboRequest, res: Response) => {
    const userEmail = req.requestData.decoded.email;
    const recipeId = req.params.id;

    this._imgService.persistImage(req.file.fieldname, DataType.RECIPE_IMAGE, userEmail)
      .then((url: string) => this._getRecipe(recipeId, url))
      .then((obj: any) => this._addImage(obj.recipe, obj.url))
      .then((recipe: Recipe) => this._db.recipes.update(recipeId, recipe))
      .then(() => this.getById(req, res))
  }

  private _getRecipe = (recipeId: number, url: string): Promise<any> => {
    return this._db.recipes.getById(recipeId).then(recipe => Promise.resolve({recipe, url}));
  }

  private _addImage = (recipe: Recipe, url: string): Promise<any> => {
    recipe.addImageUrl(url);
    return Promise.resolve(recipe);
  }
```

it's not tested, because i don't know to send an image POST request.

server/src/logic/handler/routes/recipe.handler.ts
```diff
@@ -1,17 +1,22 @@
...
 
 export class RecipeHandler {
 
   private _db: IDatabase<DbExtensions>&DbExtensions;
   private _recipeMiddleware: RecipeMiddleware;
+  private _imgService: ImageService;
 
-  constructor (db: any, recipeMiddleware: RecipeMiddleware) {
+  constructor (db: any, recipeMiddleware: RecipeMiddleware,
+               imgService: ImageService) {
```
flashback2k14: this new line is not needed. :-)

server/src/logic/middleware/recipe.middleware.ts
```diff
@@ -1,27 +1,52 @@
...
 
-  constructor (db: any, modelUtils: ModelUtils) {
+  constructor (db: any) {
     this._db = db;
-    this._modelUtils = modelUtils;
   }
 
-  public updateRecipe = (req: Request): Promise<Recipe> => {
-    const newRecipe: any = req.body;
+  public newUpdateRecipes = (req: Request, res: Response): Promise<Recipe> => {
```
flashback2k14: why have you added this? i don't get it. the modelUtils looks much better.

server/src/logic/services/filesystem-image.service.ts
hetec: I just forgot to move it out of the promise. I will do this.

server/README.md
hetec: There are no Body or Headers needed

server/package.json
hetec: OK

server/src/logic/handler/routes/recipe.handler.ts
hetec: Will remove it

server/src/config.ts
hetec: OK

server/src/logic/handler/routes/recipe.handler.ts
hetec: Looks much better I will adopt this! Thanx

server/src/logic/middleware/recipe.middleware.ts
hetec: With the model utils approach there was always a empty fields property in the response. I tried to remove it from the response but this had broken the update functionality. Maybe I find a way to keep the model utils way.

server/README.md
flashback2k14: Okay. Then I need to google how to send the request. :-)

server/src/logic/middleware/recipe.middleware.ts
flashback2k14:
```ts
delete transformedResult.fields;
```
Doesn't fix this inside _transformResult()?

server/src/logic/services/filesystem-image.service.ts
flashback2k14: Okay. No Problem.

server/src/server.ts
hetec: Yes that is ok

hetec:
To add a image via Postman go to body > form-data > name the key recipeImage > choose file instead of text (little drop down in the key field)

flashback2k14:
Ahh. Thanks. +1

server/src/logic/middleware/recipe.middleware.ts
hetec: no, if I use this the model utils will break, since the model utils functions relies on a response which contains a fields property

server/README.md
hetec: See comment at the end

flashback2k14:
works like a charm! 👍  

three more changes needed: 
- package.json --> create file folders before start server
```json
  "version": "0.0.11",
  ...
  "start:p": "npm-run-all create:l tsc --parallel serve open:b",
  "start:d": "npm-run-all create:l tsc --parallel tsc:w serve:w",
  ...
  "create:f": "mkdir -p tmp socobo_data",
  ...
  "clean:f": "rm -rf tmp/ socobo_data/",
  "clean:a": "rm -rf dist/ node_modules/ logs/ tmp/ socobo_data/",
  ...
```
- tmp file was doesn't deleted after move to socobo_data
- ERRORS.IMAGE_UPLOAD has a placeholder, but it's not set --> with this solution it's not possible to set the placeholder at the specific error places (inside _createDirIfNotExists(..), etc.)

flashback2k14:
okay. it only works if the folders are create before the server starts. maybe we should test node 8 and [util-promisify](http://2ality.com/2017/05/util-promisify.html)!?

hetec:
Yes util-promisify looks very handy!
So we need to create the folders by hand in beforehand?
I remove the placeholder in the error message

flashback2k14:
Not all only tmp and socobo_data. And this makes a npm script. :-) yes, remove this.

Issue #77, opened 2017-06-28
socobouser new db schema


flashback2k14:
ToDo:
- extend & modify
- [x] model classes
- [x] database functionality

Issue #75, opened 2017-06-20
extend database functionality


flashback2k14:
this is the main issue for following tasks:
- [x] update database schema (simplify or restructure)  [Link](https://drive.google.com/file/d/0B_-yrSSE644XX1cxT0hPR3BqUnc/view?usp=sharing) [Diagram](https://www.draw.io/?state=%7B%22ids%22:%5B%220B_-yrSSE644XX1cxT0hPR3BqUnc%22%5D,%22action%22:%22open%22,%22userId%22:%22103183073414466549375%22%7D#G0B_-yrSSE644XX1cxT0hPR3BqUnc  )
- [x] extend model classes
- [x] create new model classes
- [x] extend api 
- [x] extend routes

Related Issues:

- [x] #76 Recipe Steps 
- [x] #77 SocoboUser New DB schema

branch __75-extend-database-functionality__ is the base branch for the sub tasks.

hetec:
How do you want to simplify the database model? Do you want to remove some functions like sharing data?
The more I look at this database model, the more I get the feeling that the nosql database model of the first Socobo Project was somehow simpler.

hetec:
I think we should organize the subtasks around parts of the database model like ingredients, units or the steps for the recipes, so that we get a structure like this:
ingredients:
- endpoints
- database service
- model classes
- routes

unit
- endpoints
- model classes
- database service
- routes

....

flashback2k14:
Not the model is the problem. More the diagram. Maybe we should write down the relations like I have started!? Yes, maybe GraphQL is a better solution, but so much time spending to deal with postgres.

flashback2k14:
The new project structure sounds good. All on one place. +1

hetec:
No idea how GraphQL works, is it a full replacement of postgres? Ist it a database or just a query language?
Yes u are right, we have already invested  a lot in all this postgres stuff, and we would need to reimplement the existing db layer if we switch to another solution.

hetec:
Maybe we should create two diagrams. One which only contains the entities with their fields and no relations. And one which contains only relations and entity names (and if needed a textual description)

flashback2k14:
Hmm. If I read it right than is it only a query language. We should definitly stay with our solution. 

Yes, that sounds great. we should do this. :-)

flashback2k14:
@hetec could you please verify, modify or extend my newly created database diagram (link above)!?

hetec:
[...]

PR #79, opened 2017-07-17
#77 - socobouser new db schema

flashback2k14:
SORRY for that huge PR!!!!

this PR contains the following:
- added:
  - SQL scripts for creation SocoboUserImage, SocoboUserProvider & SocoboUserRole Tables
  - model classes for SocoboUserImage, SocoboUserProvider & SocoboUserRole
  - repository, handler & route classes for SocoboUserImage, SocoboUserProvider & SocoboUserRole
- modify:
  - the whole application structure --> structuring into domains like your suggestions from issue #75


hetec:
Hi, I'm also not able to fix the travis problem :( But I think the Problem is the use of the node types you have added to the package.json (version 8). The node version defined in the travis.yml is version 7. Where do you use these type definitions?

server/package.json
```diff
@@ -54,6 +54,7 @@
     "@types/jsonwebtoken": "~7.1.33",
     "@types/mocha": "~2.2.33",
     "@types/multer": "0.0.33",
+    "@types/node": "~8.0.12",
```
hetec: This maybe conflicts with the version in the travis.yml

server/sql/dev/create-tables-dev.sql
```diff
@@ -1,18 +1,47 @@
...
+
+CREATE TABLE Socobo_User_Role (
+  id SERIAL PRIMARY KEY,
+  name VARCHAR NOT NULL UNIQUE,
+  created NUMERIC NOT NULL,
+  lastModified NUMERIC NOT NULL CHECK (lastModified >= created)
+);
+
+CREATE TABLE Socobo_User_Provider (
+  id SERIAL PRIMARY KEY,
```
hetec: I think it would be better to use the name (is already unique) as id. Else we always need to look into the db to figure out the current id. If we use the names (should be constant) as id we could always rely on the chosen name. We could also use the name directly in the frontend. If we use the id we would alway need to do a request for the providers to get the ids.

server/sql/dev/create-tables-dev.sql
```diff
@@ -1,18 +1,47 @@
 \c socobo_dev;
 
+--
+--  SOCOBO USER
+--
+
+CREATE TABLE Socobo_User_Role (
```
hetec: Same as below

server/sql/dev/create-tables-dev.sql
```diff
@@ -1,18 +1,47 @@
...
+);
+
+CREATE TABLE Socobo_User_Provider (
+  id SERIAL PRIMARY KEY,
+  name VARCHAR NOT NULL UNIQUE,
+  created NUMERIC NOT NULL,
+  lastModified NUMERIC NOT NULL CHECK (lastModified >= created)
+);
+
+CREATE TABLE Socobo_User_Image (
```
hetec: How do we know the id of the user image? The user isn't registered at this moment and never has added a image. If you want to refer to a default image in this case I would prefer to define this url as a constant in the config because if we want to change the image we would need to change it in the db what would also change the id. So we would also need to do a request to figure out the current id of the default image.

server/src/app/repositories/Base.repository.ts
```diff
@@ -0,0 +1,75 @@
+import { IDatabase } from "pg-promise";
+import {
+  DbExtensions, ERRORS, ErrorUtils, IBaseRepositiory
+} from "../index";
+
+export class BaseRepository <T> implements IBaseRepositiory <T> {
```
hetec: I'm not sure if this is a bit to much at this point because I'm not able to anticipate how often we really could use it. But we will see.

server/src/config.ts
```diff
@@ -13,7 +11,7 @@ export class Config {
   public static TOKEN_EXPIRATION: string = "1d";
   public static TOKEN_ISSUER: string = "socobo";
   public static TOKEN_HEADER: string = "x-access-token";
-  public static DEFAULT_USER_IMAGE: string = "http://placehold.it/350x150";
+  public static DEFAULT_USER_IMAGE_ID: number = 1;
```
hetec: I think we should not rely on the database id since it is much harder to change the url of the image. I liked the url in the config file and would not mix the default image with user images. What do you think is the benefit of storing the url in the db?

server/package.json
flashback2k14: change node version inside travis.yml doesn't fix the problem! :-(

server/src/app/repositories/Base.repository.ts
flashback2k14: I use this 3 times. :-)

server/sql/dev/create-tables-dev.sql
flashback2k14: hmm. i'm not really sure if i like strings as id's. we could return id and name as key value pairs to the frontend!?

server/sql/dev/create-tables-dev.sql
flashback2k14: > How do we know the id of the user image? The user isn't registered at this moment and never has added a image.

a new user has always the default user image (=== profile pic) which is stored inside our db. if the user like he could change this and upload a new image with a new id and this id would update socobo_user. 

> If you want to refer to a default image in this case I would prefer to define this url as a constant in the config because if we want to change the image we would need to change it in the db what would also change the id.

we could change the default with a simple sql update. this  doesn't update the id.

> So we would also need to do a request to figure out the current id of the default image.

default id is stored inside the Config class

server/src/app/models/BaseObject.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export class BaseObject {
+
+  @IsNotEmpty()
+  public created: number;
+
+  @IsNotEmpty()
+  public lastModified: number;
+
+  public addCreated = (created: number): this => {
```
flashback2k14: okay. i change it to setXXX instead of addXXX.

server/src/config.ts
flashback2k14: > I think we should not rely on the database id since it is much harder to change the url of the image.

a simple sql update change the image url.

>  I liked the url in the config file and would not mix the default image with user images. What do you think is the benefit of storing the url in the db?

okay, i think we have different meaning about user images. For me user image === profile pic!?

server/src/socobouser/handlers/SocoboUserImages.handler.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export class SocoboUserImagesHandler {
+
+  private _db: IDatabase<DbExtensions>&DbExtensions;
+
+  constructor (db: any) {
+    this._db = db;
+  }
+
+  public getAll = (req: Request, res: Response): void => {
```
hetec: Is it possible to have multiple user images?

server/src/socobouser/handlers/SocoboUserImages.handler.ts
```diff
@@ -0,0 +1,41 @@
...
+      .catch((e: any) => res.status(e.statusCode).json(e.forResponse()));
+  }
+
+  public deleteById = (req: Request, res: Response): void => {
+    const imageId: number = +req.params.id;
+    this._db.socobouserImages.deleteById(imageId)
+      .then((result: Object) => res.status(200).json(result))
+      .catch((e: any) => res.status(e.statusCode).json(e.forResponse()));
+  }
+}
```
hetec: How is the user image saved?

server/src/socobouser/handlers/SocoboUserRoles.handler.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  public getById = (req: Request, res: Response): void => {
+    const roleId: number = +req.params.id;
+    this._db.socobouserRoles.getById(roleId)
+      .then((result: SocoboUserRole) => res.status(200).json(result))
+      .catch((e: any) => res.status(e.statusCode).json(e.forResponse()));
+  }
+
+  public updateById = (req: Request, res: Response): void => {
```
hetec: Do we really need to update roles programmatically from the app? I thought this would be a fixed set of roles.

server/src/socobouser/handlers/SocoboUserRoles.handler.ts
```diff
@@ -0,0 +1,40 @@
...
+    const roleId: number = +req.params.id;
+    const fieldValue: string = req.body.fieldValue;
+    this._db.socobouserRoles.updateById(roleId, fieldValue)
+      .then((result: SocoboUserRole) => res.status(200).json(result))
+      .catch((e: any) => res.status(e.statusCode).json(e.forResponse()));
+  }
+
+  public deleteById = (req: Request, res: Response): void => {
```
hetec: Should we really be able to delete roles from the app? If we delete roles does this not effect the code where the role will be checked?

server/package.json
flashback2k14: i have found the solution [source](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/16261). process.env is only with string literals instead of dottes accessible since typescript 2.2.

server/src/socobouser/handlers/SocoboUserRoles.handler.ts
hetec: Where do we use the roles values from the database? I'm not sure if I understand the purpose of storing roles in the db. Couldn't we simply store the name of the role as String into the user relation?

server/src/socobouser/routes/SocoboUserProviders.route.ts
```diff
@@ -0,0 +1,36 @@
...
+    this._router.get("/:id",
+      this._authValidationHandler.checkToken,
+      this._socoboUserProvidersHandler.getById);
+
+    this._router.put("/:id",
+      this._authValidationHandler.checkToken,
+      this._modelValidationHandler.validate(SocoboUserProvider, [ValidationGroup.USER]),
+      this._socoboUserProvidersHandler.updateById);
+
+    this._router.delete("/:id",
```
hetec: I think providers should also be constant during the runtime of the app or why do you want to be able to delete a provider?

delete and update should only be possible for admins

server/src/socobouser/routes/SocoboUserRoles.route.ts
```diff
@@ -0,0 +1,36 @@
...
+  public createRoutes (): Router {
+    this._router.get("/",
+      this._authValidationHandler.checkToken,
+      this._socoboUserRolesHandler.getAll);
+
+    this._router.get("/:id",
+      this._authValidationHandler.checkToken,
+      this._socoboUserRolesHandler.getById);
+
+    this._router.put("/:id",
```
hetec: Delete and Update only for admins. Same as providers -> I'm not sure why we would like to delete roles

server/src/socobouser/routes/SocoboUsers.route.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  constructor (
+    private _router: Router,
+    private _socoboUserHandler: SocoboUserHandler,
+    private _authValidationHandler: AuthValidationHandler,
+    private _modelValidationHandler: ModelValidationHandler
+  ) {}
+
+  public createRoutes (): Router {
+    this._router.get("/",
```
hetec: I think get all users should also only be available for admins

server/src/socobouser/handlers/SocoboUser.handler.ts
```diff
@@ -13,30 +13,30 @@ export class UserHandler {
...
-    const updateType: UpdateType = req.body.updateType;
+    const updateType: SocoboUserUpdateTypes = req.body.updateType;
     const fieldValues: string[] = req.body.fieldValues;
-    this._db.users.updateById(userId, updateType, fieldValues)
+    this._db.socobousers.updateById(userId, updateType, fieldValues)
       .then((result: SocoboUser) => res.status(200).json(result))
       .catch((e: any) => res.status(e.statusCode).json(e.forResponse()));
   }
 
   public deleteById = (req: Request, res: Response): void => {
```
hetec: I think for update and delete we need to check if the id of the current user is the same as the id of the user who should be updated because else it would be possible to change other users (if I know the id), or I'm, wrong?

server/src/config.ts
hetec: I also think of this as profile pics but I think the default image is some kind of app data and the user profile pic is user data and I'm not sure if it is a good idea to mix them up.

server/sql/dev/create-tables-dev.sql
hetec: I think my main problem is that I don't really get the point of storing this kind of things (roles, providers) to the db? What is the benefit? Whats the problem with simply using a enum for roles and providers and checking against these values?

Do we really use the values from the db or do we only check if the id in the user entity has a certain value?

hetec:
I tried to register with:

{
	"email": "patrick.hebner@user.com",
	"username": "patrick",
	"password": "password",
	"socoboUserRoleId": 0,
	"socoboUserImageId": 0,
	"socoboUserProviderId": 0
}

and got an internal server error. 

Could you update the Readme so I know which values I need to use?

Why do I need to specify all these values couldn't we set the correct default values if they are not present in the request?

server/sql/dev/create-tables-dev.sql
flashback2k14: i simply implemented our database schema. maybe we don't need extra tables for role and provider. This stuff is only admin related.

server/src/socobouser/handlers/SocoboUserImages.handler.ts
flashback2k14: nope. one image per user.

server/src/socobouser/handlers/SocoboUserImages.handler.ts
flashback2k14: hmm. looks like i forgot to implement the file-service call.

server/src/socobouser/handlers/SocoboUserRoles.handler.ts
flashback2k14: maybe an admin related stuff!?

server/src/socobouser/routes/SocoboUserProviders.route.ts
flashback2k14: yes, admin related stuff

server/src/socobouser/routes/SocoboUserRoles.route.ts
flashback2k14: yes, admin related stuff

server/src/socobouser/routes/SocoboUsers.route.ts
flashback2k14: you are right.

server/src/socobouser/handlers/SocoboUser.handler.ts
flashback2k14: you are right. this is a needed check!

flashback2k14:
Zero id values doesn't exist. use one.

flashback2k14:
what should i change? 
- delete role and provider tables, repos, handlers and routes? 
- add saving images
- update readme

flashback2k14:
@hetec what did you think about the new structure?

server/src/socobouser/handlers/SocoboUserRoles.handler.ts
hetec: I think we will define the roles once at development time (an can extend the roles if we update the app). I think the admin should maybe able to give or remove a role from a user but he should not be able to create new roles. What do you think.

hetec:
Its hard to say because you have invested a lot of work into the roles and providers. But I think we would need a lot less code if we only create a enum (or some kind of constants files) for the roles and providers since we have very few of them. But I leave the decision to you!

Adding image is necessary in my opinion and updating the readme would be great.

The new structure is very nice. I like it.

Thanks for your work!

hetec:
Shame on me ;) Zero id values was really foolish

flashback2k14:
Okay. I add some role checks, implement the image saving logic and update the readme.

server/src/socobouser/handlers/SocoboUser.handler.ts
flashback2k14: @hetec how would you implement this? i can't get a solution. :-/

flashback2k14:
@hetec i have implemented the following things
- role checks
- socobouserimage saving logic 
- update the readme

Issue #85, opened 2017-08-20
mongodb socobouser


flashback2k14:
**todo's:**
  - [x] collection / model class
  - [x] repository class
  - [x] handler class
  - [x] route class
  - [x] remove obsolete classes 
  - [x] update readme
  - [x] testing

Issue #80, opened 2017-08-04
change to MongoDB


flashback2k14:
- database schema  [Link](https://drive.google.com/file/d/0B_-yrSSE644XX1cxT0hPR3BqUnc/view?usp=sharing) [Diagram](https://www.draw.io/?state=%7B%22ids%22:%5B%220B_-yrSSE644XX1cxT0hPR3BqUnc%22%5D,%22action%22:%22open%22,%22userId%22:%22103183073414466549375%22%7D#G0B_-yrSSE644XX1cxT0hPR3BqUnc)
- mongoDB Relationship [Link](https://docs.mongodb.com/manual/tutorial/model-referenced-one-to-many-relationships-between-documents/)
- mongoDB Tutorial [Link](https://www.tutorialspoint.com/mongodb/index.htm)
- useful library
  - https://github.com/szokodiakos/typegoose
  - http://codingsans.com/blog/mongoose-models-using-typescript-classes

DEV Branch: **80-change-to-MongoDB** create from Branch **75-extend-database-functionality**
- ToDo's:
  - [x] database setup #84 
  - [x] socobouser collection + repository #85 
  - [x] recipe collection + repository #87 
  - [x] fooditem collection + repository #90

flashback2k14:
@hetec i have translated our postgres database schema into a mongoDB schema. Please take a look. i'm not sure if all is right. thanks.

hetec:
@flashback2k14 Some ideas to the new schema. I would maybe embed the recipeSteps, recipeMetaData and the recipe Images directly into the recipe document?! Because each recipe always has its unique steps and images we will not duplicate data if we include them into the recipes. On the other hand we don't need to do separate db queries to fetch the data. To have a own collection for the categories is ok since different recipes can use the same categories.

The fooditem stuff looks very fine for me!

flashback2k14:
@hetec i follow the provided mongoDB pattern for [Model One-to-Many Relationships with Document References](https://docs.mongodb.com/manual/tutorial/model-referenced-one-to-many-relationships-between-documents/)

> To avoid mutable, growing arrays, store the publisher reference inside the book document:

maybe i have misunderstood it. Could you please check the link!?

hetec:
The recipe ingredients and recipe user stuff looks also fine

flashback2k14:
okay the _maybe parts_ could integrated into the schema?

hetec:
Yes I think they are 👍

[...]

PR #93, opened 2017-09-02
#87 mongodb recipes

hetec:
- adapt repository to use mongodb 
- remove resipe-steps repository and integrate steps into recipe model 
- use another  middleware approache to update single fields, no separate endpoint 
- adapt handlers to work with the new repository - adapt the mongo db setup to consider the recipe repository

Based on #85


flashback2k14:
This PR goes into master!? The "master" branch for the mongodb changes is #80 or #85.

hetec:
Sorry, I thought that this PR would first be merged after #85 has already been merged into master.

server/sql/dev/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
+version: '3'
```
flashback2k14: how could i run this docker container. I have zero knowledge about docker? what's the connectionstring? is it possible to add the "docker tools" as devDedenpency and create a npm-script that starts the docker container?

server/README.md
```diff
@@ -506,16 +448,15 @@ npm run lint
...
 
   Alters the properties, defined in the request body, on the recipe with the
-  provided id. The request body must contain all recipe properties except the id.
-  To update only specific field you need to define these fields in the fields query parameter.
-  If a fields query prameter is defined, only these fields will be updated.
+  provided id.
+  If the request body only conatains single fields of a recipe, only these fields
+  will be updated. This is true for the first level of properties. E.g. if you update
+  the steps of a recipe and you only provide a single step all remaining steps are
+  removed since there are no data in the request.
```
flashback2k14: > if you update the steps of a recipe and you only provide a single step all remaining steps are removed since there are no data in the request.

what if we use the stepNumber as identification? or add an objectId? That's makes it possible to update a single recipe step!? what did you think?

server/src/recipe/handler/recipe.handler.ts
```diff
@@ -1,150 +1,122 @@
 import { NextFunction, Request, Response } from "express";
-// import { IDatabase } from "pg-promise";
-// import {
-//   ApiError, DataType, DbExtensions, ERRORS, ImageService, SocoboRequest
-// } from "../../app/index";
+import { Types } from "mongoose";
```
flashback2k14: this is obsolete.

server/src/recipe/handler/recipe.handler.ts
```diff
@@ -1,150 +1,122 @@
...
+    try {
+      const result = await this._db.recipe.getAll();
+      res.status(200).json(result);
+    } catch (error) {
+      this._sendError(res)(error);
     }
   }
 
-  public searchByField = (req: Request, res: Response): void => {
+  public searchByField = async (req: Request, res: Response) => {
```
flashback2k14: should we also add searchByFields?

server/test/5-recipe.route.test.ts
```diff
@@ -1,306 +0,0 @@
-/*tslint:disable:no-unused-expression*/
-
-process.env["NODE_ENV"] = "test";
```
flashback2k14: would you add new tests into this PR or another?

server/tslint.json
```diff
@@ -8,6 +8,7 @@
     "ordered-imports": [true, { "named-imports-order": "case-insensitive" }],
     "member-ordering": [true, { "order": [ "constructor" ] }],
     "interface-name": false,
-    "no-string-literal": false
+    "no-string-literal": false,
+    "prefer-object-spread": false
```
flashback2k14: why you add this?

hetec: If prefer-object-spread is true I cannot use Object.assign in the way I did to merge object. This would lead to an lint error

server/test/5-recipe.route.test.ts
hetec: In another I think

server/src/recipe/handler/recipe.handler.ts
hetec: yes sure

server/README.md
hetec: There is already a _id for recipe steps. I means it is not possible with the current implementation. And I wanted to see if we really need to update single steps in the future before I implement this.

server/src/recipe/handler/recipe.handler.ts
hetec: Recipe already had this feature. We don't need to implement this in each domain but I think it makes sense to search in recipes. What do u think?

server/sql/dev/docker-compose.yml
hetec: I only added this to be able to use the mongo db inside a docker container. Because I didn't want to install mongo db locally. So you don't really need to use this if you don't want. But if you want to try it I can add some notes to the readme

flashback2k14: yes, i'm interested. please add some notes. thanks.

server/README.md
flashback2k14: Okay.

server/src/recipe/handler/recipe.handler.ts
flashback2k14: > Recipe already had this feature. We don't need to implement this in each domain
i can't find it? what did you mean?

> I think it makes sense to search in recipes. What do u think?
yes i think this makes sense.

server/tslint.json
flashback2k14: you could use the spread operator 
```js
{...recipe,...requestBody}
```
instead of 
```js
Object.assign(recipe, requestBody);
```

hetec: Yes I tried it but it didn't work. I always need to something like

const variable = {...recipe,...requestBody}
const realData = variable._doc

Else I would get a object with a lot of meta data like cache things and so on.

Do you know why I do not get a simple Object after using spread?

flashback2k14:
Recipe is a document&recipe with alle properties from document and recipe. This is why I added the transform function to get only a socobouser. What if you use 
```js
{...recipe._doc, ...requestBody}
```
?

flashback2k14:
The spread operator merge all properties of all object and recipe has many properties.

hetec:
Ok, so the spread operator also merges the not 'own' properties of the prototype chain. Good to know.

This seams to work. One Problem is that I always get the type error that _doc is not a property of the recipe, but it exists and can be used.

flashback2k14:
Maybe a transform function to create a real recipe object is a good idea or simple use the object.Assign. the use of _doc is not a good idea, because it's a private property from the document class.

hetec:
I think I could resolve this using the lean() function of mongoose. Therefor mongoose only returns the plain object without the whole bunch of meta information.

server/README.md
```diff
@@ -54,6 +54,31 @@
...
+
+# Clean up your containers
+npm run docker:rm
+```
+
+4. To add new mock data, add a new mongodb-seed section in the docker-compose file
+
+in your terminal
+
+6.
```
flashback2k14: i think here is something missing!? :-)

Issue #94, opened 2017-09-03
Extend Recipe Model


hetec:
- Extend recipe image section
- Extend recipe with meta data

PR #98, opened 2017-09-21
#94 extend recipe model

hetec:
- level property
- duration property
- multiple images for one recipe
- delete images
- tests


server/sql/automate-creation.sh
```diff
@@ -4,7 +4,7 @@ if [ "$1" = "-d" ]
 then
 
   mongoimport --db socobo_dev --collection socobouser --drop --file ./data/create-socobouser.json --jsonArray
-  # TODO: Add Demo Data: mongoimport --db socobo_dev --collection recipe --drop --file ./dev/create-recipe.json
+  mongoimport --db socobo_dev --collection recipes --drop --file ./data/create-recipe.json --jsonArray
```
flashback2k14: Can we name our collections singularly? It's a collection of recipe or it's a collection of socobouser.

server/sql/dev/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+      - "27017:27017"
+
+  mongo-seed-user:
+    image: mongo
+    links:
+      - mongodb
+    volumes:
+      - ./:/seed-data
+    command:
+      'bash -c "mongoimport --host mongodb --db socobo_dev --collection socobousers --type json --file /seed-data/create-socoboUser.json --jsonArray"'
```
flashback2k14: i think the socobouser collection 'socobousers' should fail, because have named it 'socobouser' everywhere.

server/src/recipe/handler/recipe.handler.ts
```diff
@@ -70,7 +81,9 @@ export class RecipeHandler {
   }
 
   public update = async (req: Request, res: Response) => {
-    const recipe: Recipe = new Recipe().clone(req.body as Recipe);
+    let recipe: Recipe = new Recipe().clone(req.body as Recipe);
```
flashback2k14: if you provide a deleteImage Function into the recipe model object you could use ```const``` again.

server/src/recipe/handler/recipe.handler.ts
```diff
@@ -107,16 +120,37 @@ export class RecipeHandler {
   public uploadImage = async (req: SocoboRequest, res: Response) => {
     const userEmail = req.requestData.decoded.email;
     const recipeId = req.params.id;
+    const imageTitle = req.query.title;
+    if (!imageTitle || !req.file.filename) {
```
flashback2k14: should we integrate the recipe title check into a middleware?

server/src/recipe/handler/recipe.handler.ts
```diff
@@ -1,21 +1,31 @@
+import { validate } from "class-validator";
```
flashback2k14: Is this import needed?

server/src/recipe/handler/recipe.handler.ts
```diff
@@ -1,21 +1,31 @@
...
 import { NextFunction, Request, Response } from "express";
+import { Types } from "mongoose";
```
flashback2k14: Is this import needed?

server/src/recipe/models/recipe.ts
```diff
@@ -69,13 +104,23 @@ export class Recipe implements Validatable {
     return this;
   }
 
-  public setImageUrl (imageUrl: string) {
-    this.imageUrl = imageUrl;
+  public setImageUrl (images: RecipeImage[]) {
```
flashback2k14: maybe rename this to setImageUrls or setImages!?

server/src/recipe/validators/recipe-level.validator.ts
```diff
@@ -0,0 +1,42 @@
+import {
+  registerDecorator, ValidationArguments, ValidationOptions,
+  ValidatorConstraint, ValidatorConstraintInterface
+} from "class-validator";
+import { Level, Recipe } from "../index";
+
+@ValidatorConstraint({ name: "RecipeLevel", async: false })
+export class RecipeLevel implements ValidatorConstraintInterface {
```
flashback2k14: Could you please rename this class to RecipeLevelValidator?

server/sql/automate-creation.sh
hetec: yes we can, but usually it's a collection of socobousers since there are multiple users in one collection.

server/sql/dev/docker-compose.yml
hetec: Oh the file in /dev is the wrong one. You need to use the one directly in sql. There I use socobouser as collection. I remove the dev folder I think we don't need it anymore since all data are hold in /data

server/src/recipe/handler/recipe.handler.ts
hetec: no, thanks

server/src/recipe/handler/recipe.handler.ts
hetec: Nice tip!

hetec:
@flashback2k14 Thanks for your hints! I hope I have resolved all issues.

Issue #96, opened 2017-09-20
mongodb fooditem category


flashback2k14:
- [x] FoodItemCategory
  ![bildschirmfoto 2017-10-08 um 10 41 17](https://user-images.githubusercontent.com/6609359/31315283-68e0d694-ac15-11e7-926e-294a1dcb5868.png)
- [x] testing
- [x] readme

PR #105, opened 2017-12-15
#96 - mongodb fooditem category

flashback2k14:
implement Base Repository to reduce duplicated code

### DIFF
--- a/server/src/food/handler/FoodItemCategory.handler.ts
+++ b/server/src/food/handler/FoodItemCategory.handler.ts
@@ -1,0 +1,66 @@
+import { Request, Response } from "express";
+import { Types } from "mongoose";
+import { DbExtension } from "../../db/interface/db-extension";
+import { FoodItemCategory } from "../index";
+
+export class FoodItemCategoryHandler {
+
+  constructor (private _db: DbExtension) {}
+
+  public getAll = async (req: Request, res: Response): Promise<void> => {
+    try {
+      const result = await this._db.fooditemCategory.getAll();
+      res.status(200).json(result);
+    } catch (error) {
+      res.status(error.statusCode).json(error.forResponse());
+    }
+  }
+
+  public getById = async (req: Request, res: Response): Promise<void> => {
+    try {
+      const id = new Types.ObjectId(req.params.id);
+      const result = await this._db.fooditemCategory.getById(id);
+      res.status(200).json(result);
+    } catch (error) {
+      res.status(error.statusCode).json(error.forResponse());
+    }
+  }
+
+  public save = async (req: Request, res: Response): Promise<void> => {
+    try {
+      const category = new FoodItemCategory().clone(req.body);
+
+      // TODO:  await this._db.fooditem.getById(category.foodItemId);
+
+      const result = await this._db.fooditemCategory.save(category) as Types.ObjectId;
+      res.status(201).json(category.setId(result));
+    } catch (error) {
+      res.status(error.statusCode).json(error.forResponse());
+    }
+  }
+
+  public updateById = async (req: Request, res: Response): Promise<void> => {
+    try {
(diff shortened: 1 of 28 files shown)


server/src/food/handler/FoodItemCategory.handler.ts
```diff
@@ -0,0 +1,66 @@
...
+      const result = await this._db.fooditemCategory.getAll();
+      res.status(200).json(result);
+    } catch (error) {
+      res.status(error.statusCode).json(error.forResponse());
+    }
+  }
+
+  public getById = async (req: Request, res: Response): Promise<void> => {
+    try {
+      const id = new Types.ObjectId(req.params.id);
```
hetec: It's not necessary to create a ObjectId Instance your can simply use the string

server/src/food/handler/FoodItemCategory.handler.ts
```diff
@@ -0,0 +1,66 @@
...
+    } catch (error) {
+      res.status(error.statusCode).json(error.forResponse());
+    }
+  }
+
+  public save = async (req: Request, res: Response): Promise<void> => {
+    try {
+      const category = new FoodItemCategory().clone(req.body);
+
+      // TODO:  await this._db.fooditem.getById(category.foodItemId);
```
hetec: What's about this todo?

server/src/food/handler/FoodItemCategory.handler.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  public updateById = async (req: Request, res: Response): Promise<void> => {
+    try {
+      const categoryId = new Types.ObjectId(req.params.id);
+      const foodItemId = req.body.foodItemId;
+      const updatedCategoryName = req.body.name;
+
+      // TODO:  await this._db.fooditem.getById(foodItemId);
```
hetec: same as above

server/src/food/routes/FoodItemCategory.route.ts
```diff
@@ -0,0 +1,44 @@
...
+    this._router.get("/",
+      this._authValidationHandler.checkToken,
+      this._fooditemCategoryHandler.getAll);
+
+    this._router.get("/:id",
+      this._authValidationHandler.checkToken,
+      this._fooditemCategoryHandler.getById);
+
+    this._router.put("/:id",
+      this._authValidationHandler.checkToken,
```
hetec: Is everybody allowed to create, update and delete categories? Maybe it's better to provide the users with a consistent predefined set of categories and only allow admins to update this list. What do you think?

server/src/food/handler/FoodItemCategory.handler.ts
flashback2k14: yes, i know but i have always used ObjectId as type.

server/src/food/handler/FoodItemCategory.handler.ts
flashback2k14: it's not implemented at this time. that's the next step.

server/src/food/handler/FoodItemCategory.handler.ts
flashback2k14: it's not implemented at this time. that's the next step.

server/src/food/routes/FoodItemCategory.route.ts
flashback2k14: that's a good idea. i will add this check.

flashback2k14: what is with fooditem unit and fooditem template? should we restrict this too?

hetec: I'm not sure but at the moment I would say no. Let's keep this in mind

Issue #99, opened 2017-09-21
Recipe Category


hetec:
Routes:
- [x] GET recipe-categories => All categories (predefined set)
- [x] POST recipe-categories => create a new one (only Admins, unique values)

- [x] Set up new model for recipe-categories
- [x] Add reference to recipe

![image](https://user-images.githubusercontent.com/6704069/30715644-5bfde332-9f17-11e7-9e41-f3072e6e36f1.png)

Correction to the image above => the ref goes into the recipe document

PR #106, opened 2017-12-23
#99 - recipe category

hetec:
recipe category

- endpoints
- db access
- handlers
- tests


flashback2k14:
bildschirmfoto 2017-12-25 um 11 27 38
could you please update your docker description with this PR!?

flashback2k14:
bildschirmfoto 2017-12-25 um 11 29 51
i recommend to use the tslint plugin. it saves many time. :-)

server/src/recipe/handler/recipe.handler.ts
```diff
@@ -19,19 +20,63 @@ export class RecipeHandler {
     };
   }
 
+  private _resolveCategory = (recipe: Recipe): Promise<Recipe> => {
```
flashback2k14:
```ts
  private _resolveCategory = async (recipe: Recipe): Promise<Recipe> => {
    const result = await this._db.recipeCategories.getById(recipe.categoryId);
    recipe.category = result as RecipeCategory;
    return recipe;
  }
```
i would change it to use async await

server/src/recipe/handler/recipe.handler.ts
```diff
@@ -19,19 +20,63 @@ export class RecipeHandler {
...
+    return this._db.recipeCategories.getById(recipe.categoryId)
+      .then((category: RecipeCategory) => {
+        recipe.category = category;
+        return recipe;
+      });
+  }
+
+  private _resolveAllCategories = (recipes: Recipe[]): Promise<Recipe[]> => {
```
flashback2k14:
```ts
private _resolveAllCategories = async (recipes: Recipe[]): Promise<Recipe[]> => {
    return await Promise.all(recipes.map(async (recipe: Recipe) => {
      return await recipe.categoryId
        ? this._resolveCategory(recipe)
        : Promise.resolve(recipe);
    }));
  }
```
async await with map. looks simpler!?

server/src/recipe/handler/recipe.handler.ts
```diff
@@ -19,19 +20,63 @@ export class RecipeHandler {
...
+    recipes.forEach((recipe: Recipe) => {
+      const recipeWithCategoryPromise = recipe.categoryId
+        ? this._resolveCategory(recipe)
+        : Promise.resolve(recipe);
+      promises = [...promises, recipeWithCategoryPromise];
+    });
+    return Promise.all(promises);
+  }
+
+  private _mapRecipesToRecipesWithCategory = (recipes: Recipe[]) => {
```
flashback2k14:
```ts
private _mapRecipesToRecipesWithCategory = (recipes: Recipe[]) => {
    return recipes.map((recipe: Recipe) => {
      return new Recipe().clone(recipe).setCategoryId(undefined);
    });
  }
```

server/src/recipe/routes/recipe-category.route.ts
```diff
@@ -0,0 +1,45 @@
...
+import { ModelValidationHandler, ValidationGroup } from "../../app/index";
+import { AuthValidationHandler } from "../../auth/index";
+import { SocoboUserRoleType } from "../../socobouser/index";
+import { Recipe, RecipeCategoryHandler, RecipeMiddleware } from "../index";
+
+export class RecipeCategoryRoute {
+
+  constructor (
+    private _router: Router,
+    private _multer: Instance,
```
flashback2k14: i think this is not needed.

server/src/recipe/routes/recipe-category.route.ts
```diff
@@ -0,0 +1,45 @@
...
+    private _recipeHandler: RecipeCategoryHandler,
+    private _authValidationHandler: AuthValidationHandler,
+    private _modelValidationHandler: ModelValidationHandler,
```
flashback2k14: no model validation is implemented.

server/src/recipe/routes/recipe-category.route.ts
```diff
@@ -0,0 +1,45 @@
...
+    private _recipeMiddleware: RecipeMiddleware) {}
```
flashback2k14: i think this is not needed.

server/src/recipe/handler/recipe.handler.ts
```diff
@@ -19,19 +20,63 @@ export class RecipeHandler {
...
+      } else {
+        res.status(200).json(result);
+      }
     } catch (error) {
       this._sendError(res)(error);
     }
   }
 
   public getAll = async (req: Request, res: Response) => {
+    const queryPramas = req.query;
```
flashback2k14:
```ts
 public getAll = async (req: Request, res: Response) => {
    const queryPramas = req.query;
    try {
      const result = await this._db.recipe.getAll() as Recipe[];

      if (!queryPramas.hasOwnProperty("resolveCategory")) {
        res.status(200).json(result);
        return;
      }

      const recipesWithCategories = await this._resolveAllCategories(result);
      res.status(200).json(this._mapRecipesToRecipesWithCategory(recipesWithCategories));

    } catch (error) {
      this._sendError(res)(error);
    }
  }
```

hetec:
Thanks for your feedback, I will incorporate your improvements

server/src/recipe/handler/recipe.handler.ts
hetec: Yes this looks better. But in this case you will wait for each _resolveCategor Promise to resolve. So the requests for each category are performed sequentially. The promise only approach will perform the requests in parallel. I will see that I can use the best of both, ok?

hetec: With request I mean the db query for a category

Issue #100, opened 2017-09-21
recipe ingredient


hetec:
![image](https://user-images.githubusercontent.com/6704069/30715685-83b404c4-9f17-11e7-8f86-8dde24a47f73.png)

Relationship: many to many

modeled as:

`recipe: {
  ingredients: [id1, id2 ...]
}`

PR #107, opened 2018-01-02
#100 - recipe ingredient

hetec:
- recipe ingredient model
- recipe ingredient repository
- recipe ingredient handler
- recipe ingredient routes
- add ingredients array to recipe
- enable resolving fooditemTemplates in ingredients
- tests + readme


server/src/db/implementation/mongo-db-extension.ts
```diff
@@ -15,21 +22,32 @@ export class MongoDbExtension implements DbExtension {
...
+    private _recipeCategoryModel: Model<Document & RecipeCategory>,
+    private _recipeIngredientModel: Model<Document & RecipeIngredient>
   ) {
     this.fooditemTemplate = new FoodItemTemplateRepository(_fooditemTemplateModel);
     this.fooditemCategory = new FoodItemCategoryRepository(_fooditemCategoryModel);
     this.fooditemUnit = new FoodItemUnitRepository(_fooditemUnitModel);
     this.socobouser = new SocoboUserRepository(_socobouserModel);
-    this.recipe = new RecipeRepository(_recipeModel),
-    this.recipeCategory = new RecipeCategoryRepository(_recipeCategoryModel);
+    this.recipe = new RecipeRepository(
```
flashback2k14: this looks a little strange. RecipeRepository needs the model and the CrudRepository, because the CrudRepository doesn't provide all methods needed inside the RecipeRepository. What if your CrudRepository provide the model as a public property? but this violates the law of demeter. what did you think?

server/src/recipe/models/recipe-ingredient.ts
```diff
@@ -0,0 +1,53 @@
...
+  public fooditemTemplate: FoodItemTemplate;
+
+  public clone (recipeIngredient: RecipeIngredient) {
+    if (!recipeIngredient) {
+      return undefined;
+    }
+    this._id = recipeIngredient._id;
+    this.amount = recipeIngredient.amount;
+    this.fooditemTemplateId = recipeIngredient.fooditemTemplateId;
+    this.fooditemTemplate = new FoodItemTemplate().clone(recipeIngredient.fooditemTemplate);
```
flashback2k14: why have you the templateId and the template object?

hetec: If you use the resolve query param the response contains the the whole fooditemTemplate else it only contains the id

flashback2k14: Okay. Should we mark the object as optional?

server/src/db/implementation/mongo-db-extension.ts
hetec: Yes that is a little bit annoying, but I think it's the most flexible way and it makes the dependencies very clear. I also thought about inheritance but if I maybe need to add a method to the curdRepository class which is only needed by some classes I would break the other classes. So I decided to use composition.

I think we only have these problems because we don't use any DI thing and need to construct our object be ourself. But I think that's currently not to bad.

server/src/recipe/models/recipe-ingredient.ts
hetec: Oh, I never thought about using the object in the request => If you specify a foodItemTemplate object in the request this would be ignored (since it's not part of the schema and so on)

But yes we can mark it as optional.

server/src/db/implementation/mongo-db-extension.ts
flashback2k14: Yes, its not so Bad.

Issue #95, opened 2017-09-20
mongodb fooditem


flashback2k14:
- [x] FoodItem
  ![bildschirmfoto 2017-10-08 um 10 41 39](https://user-images.githubusercontent.com/6609359/31315292-a8e0b0fc-ac15-11e7-80f7-fb9ca940bef4.png)
- [x] extend Fooditem Template with Unit Ref Id and Category Ref Ids
- [ ] testing
- [x] readme

hetec:
yes I'm with you. The recipe ingredients had a similar error. I have put a ingredients array on the recipe instead of adding a recipe ref to the ingredient.

I currently ask myself why the categories and unit sits on the food item and not the template?! Aren't the part of the template? Example:

We have a apple template (food item of type apple 🍎 )
=> apple, cat: fruit, unit: piece, ...

And the we have multiple instances of the apple template (the concrete apples in our fridge)
=> These apples have a specific best before date and can be used for cooking (the amount will be decreased if you cook sth. with apples)

?!

PR #111, opened 2018-01-04
#112 - Proposal: Improve error handling

hetec:
This is a proposal to improve our error handling a little bit using the example of recipe.

Main goals:

- reduce boilerplate code 
=> remove a lot of try catch 
=> remove error handling methods (simply throw your error)
=> remove duplicated sending of errors with next()
- one centralized place to send error responses
- lower the risk of 'UnhandledPromiseRejection' Errors 
- remove DB Error class since it has no really benefit compared with ApiError (no additional information)


hetec:
Mmh, I have done this only for the recipe, do we want to use this kind of error handling for all modules? If yes I could implement this in the meantime your are implementing #95. An after you are ready we can merge this one?!

flashback2k14:
The PR sounds like to implement this for the whole application. If not than could we merge it.

hetec:
Ok, I will implement the remaining parts.

Issue #108, opened 2018-01-03
recipe - missing parts


flashback2k14:
- isPublic
- collaborators
- readers

![bildschirmfoto 2018-01-03 um 21 04 54](https://user-images.githubusercontent.com/6609359/34537785-0157f9cc-f0ca-11e7-94aa-17d7709234f7.png)

PR #114, opened 2018-01-13
#108 - recipe fields for sharing

patrick-hebner:
- Add fields for collaborators, readers and owner
- Add authorization methods to restrict access to recipes
- Restrict access to recipes
- Fix some box
- Add tests


server/src/app/utils/crypto.utils.ts
```diff
@@ -16,8 +16,8 @@ export class CryptoUtils {
     }
   }
 
-  public comparePasswords = async (firstPw: string, secondPw: string): Promise<void> => {
-    const result = await bcrypt.compare(firstPw, secondPw);
+  public comparePasswords = async (firstPw: string, secondPw: string, pwIsToken: boolean): Promise<void> => {
+    const result = pwIsToken ? firstPw === secondPw : await bcrypt.compare(firstPw, secondPw);
```
flashback2k14: why did you add this? i don't get your intention.

server/src/auth/handler/auth.handler.ts
```diff
@@ -12,7 +12,8 @@ export class AuthHandler {
   public login = async (req: SocoboRequest, res: Response, next: NextFunction): Promise<void> => {
     try {
       const result = await this._getExtractedRequestBodyResult(req, "login(..)");
-      const loginResult = await this._authService.login(result);
+      const pwIsToken = req.query.token;
```
flashback2k14: is `token` query param a boolean? what's the differenz if this flag is set?

server/test/helper/TestHelper.ts
```diff
@@ -39,4 +39,29 @@ export class TestHelper {
       return error;
     }
   }
+
+  public static getTokenForEmailAndPassword = async (
```
flashback2k14: is it possible to add this to the existing getToken method?

server/test/50-recipe.route.test.ts
```diff
@@ -2,20 +2,29 @@ process.env["NODE_ENV"] = "test";
 
 import { expect } from "chai";
 import * as mocha from "mocha";
+
```
flashback2k14: could you please reorder / group your tests. it costs many time if you override the token so much. :-)

flashback2k14 (commented):
LGTM except to login method extension

server/src/app/utils/crypto.utils.ts
hetec: I needed to add this since I don't know the password of the mock users. So I couldn't login with these useless to test the authorization of the recipe endpoints. If you tell me all password I can remove this again. This method allowed me to login directly with the token instead of the password.

server/src/auth/handler/auth.handler.ts
hetec: If the token query param is true I can login directly with a token instead of a password. See above comment

server/test/50-recipe.route.test.ts
hetec: No, unfortunately not. The main problem is the the test don't reset the state of the db before each test. So each test depends on the test before it since the state of the db was changed. For example I had problems to use a specific user id in the recipes because it was the id you deleted your last user test (The cause of this error was a really hard to find). May I can check if I can reorder some of them but I also need a lot of different users to test the authorization stuff.

But in general it would be a good practice to reset the db after each test. What do you think?

server/test/helper/TestHelper.ts
hetec: Maybe. I didn't want to have a method with many parameter where some of them will never be used. If both is implemented in one function we always need to set some parameters to null?! I'm not sure what's the best way here. Should I try to combine these functions to one?

server/src/app/utils/crypto.utils.ts
flashback2k14: the user passwords are inside the `30-socobouser.route.test.ts`file

please remove this. i think this is a big security issue if we could login with the password hash value.

server/src/auth/handler/auth.handler.ts
flashback2k14: this is not a token. its the password hash value you use. or get i wrong?

server/test/50-recipe.route.test.ts
flashback2k14: yes, i have the same big issues inside the fooditem tests. how could we realize the db restore after each test? i think call `npm test` with `require('child_process').exec()` is not the best solution!? maybe the restoring of the db after each test cost many time. what if we use the before or after hook?

server/test/helper/TestHelper.ts
flashback2k14: no, its okay. now i know why you use the token query param. but please remove this query param.

server/src/recipe/middleware/recipe.authorization.service.ts
```diff
@@ -0,0 +1,76 @@
+import { NextFunction, Request, Response } from "express";
+import {
+  ApiError, ERRORS, SocoboRequest
+} from "../../app/index";
+
+import { DbExtension } from "../../db/interface/db-extension";
+import { SocoboUser } from "../../socobouser/index";
+import { Recipe } from "../index";
+
+export class RecipeAuthorizationService {
```
flashback2k14: what think you about to make this as a middleware? it's not really needed, but it would looks better and you doesn't need the SocoboUserRequest inside your RecipeHandler!? :-)

server/test/50-recipe.route.test.ts
flashback2k14: we could implement the db setup programmatically inside the test folder similar i did [here](https://github.com/flashback2k14/playthatsong/blob/master/server/logic/services/setup.service.js)!?

hetec: Yes, we should do it this way and then use beforeEach and afterEach to do the setup and the cleanup?!

hetec: And I think because these tests are integration tests the speed is not the most important aspect because they are already a lot slower then unit test (because they use http an db connections and so on)

hetec: We should prefer stable and correct test over the execution time in this case I think.

server/src/auth/handler/auth.handler.ts
hetec: I just call these things token since a token is simply a hashed value. I didn't mean the auth token here.

server/src/app/utils/crypto.utils.ts
hetec: Yes I will remove this, it was only to test my routes. I didn't plan to keep this in the code.

server/test/helper/TestHelper.ts
hetec: ok

server/src/recipe/middleware/recipe.authorization.service.ts
hetec: I will check if its possible in all cases

hetec: I'm not sure how I should do this without repeating a lot of code from the handler methods. Because I need to fetch the recipe before I can check if the user is authorized. So if I want to use a middleware I would need to fetch the recipe in the middleware and then check the authorization. After this I would need to fetch the recipe again in the handler to do the remaining work. Maybe I could add the recipe / recipes to the request, but there are case where I need different recipe data in the handler (e.g. getAllRecipes and search). Any Idea?

server/test/50-recipe.route.test.ts
flashback2k14: Yes, you are right. So think the beforeEach hook should create a fresh DB. AfterEach hook it's not needed than.

server/src/recipe/middleware/recipe.authorization.service.ts
flashback2k14: Yes you could add this to the request. But your solution is good. We should use it and move on.

hetec:
So I hope it's done now

flashback2k14:
You didn't committed anything. :-)

hetec:
VS Code just didn't show me the error in the editor.

flashback2k14 (approved):

PR #115, opened 2018-01-13
#95 - mongodb fooditem

flashback2k14:
- fooditem template now contains food categories and food unit
- fooditem tests not working. i have no glue why. every time socobouser or fooditem template are not found. manual tests all working. @hetec maybe you have an idea.


hetec:
Maybe your tests fail because a previous test has already deleted the user or food item. I had a similar problem and also needed a long time to figure this out. So check if for example your user test deletes the user you want to use before your test is running. (Also see problem description in the PR comments of the recipe PR)

flashback2k14:
yes. let us discuss this inside your PR.

server/src/food/handler/FoodItem.handler.ts
```diff
@@ -0,0 +1,127 @@
...
+          ? await this._db.fooditem.getByIdResolved(id, true)
+          : await this._db.fooditem.getById(id);
+
+      res.status(200).json(result);
+    } catch (error) {
+      next(error);
+    }
+  }
+
+  public save = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
```
hetec: There was an error for me with the POST:

Request:

{
        "foodItemTemplateId": "59a2eea98a5a150e9e829409",
        "socoboUserId": "59a2ee598ae9b7b45243d503",
        "amount": 3333,
        "bestBefore": 1501948817400
    }

Error

{
    "cause": "item.socoboUserId.toHexString is not a function",
    "message": "Internal server error",
    "method": "",
    "source": "",
    "type": "TypeError"
}

server/src/food/handler/FoodItem.handler.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+  }
+
+  public updateById = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+    try {
+      const foodItemId = new Types.ObjectId(req.params.id);
+      const socoboUserId = new Types.ObjectId(req.body.socoboUserId);
+      const foodItemTemplateId = new Types.ObjectId(req.body.foodItemTemplateId);
+
+      await this._checkIfSocoboUserExists(socoboUserId.toHexString(), "updateById(..)");
```
hetec: Just a question because I forgot it or just don't know it. Why do you use a hex string here?

patrick-hebner (approved):
Looks fine! Just fix the tests and please check the error in the POST and then this one should be finished.

Maybe you could add the new db setup for the test. But this could also be done a separate PR

server/src/food/handler/FoodItem.handler.ts
flashback2k14: I will check this tomorrow I guess.

server/src/food/handler/FoodItem.handler.ts
flashback2k14: Because my inconsistent implementation. SocoboUser repository works with object id as string and toHexString(..) produce this string from Types.ObjectId. 

But if I look I doesn't need this. Thanks for pointing. I will change this too.

flashback2k14:
I add this DB test setup for the food items here.

flashback2k14:
@hetec I have fix the post are and the tests. I have added the beforeEach hooks into the food item tests.

